PR #4, opened 2016-12-21
fdtdump.c: provide colored output

xypron:
A new command line option is added
  -c, --color <arg> Colorize, argument can be auto, never or always
which defaults to auto.

For colored output in pipes use 'always', e.g.
  fdtdump -c always file.dtb | less -R

If you don't like colors use
alias fdtdump="fdtdump -c none"

Signed-off-by: Heinrich Schuchardt <xypron.glpk@gmx.de>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
+before_install:
+        - sudo apt-get -qq update
+        - sudo apt-get install -y libncurses5-dev
+
 language: c
 
 script:
-        - make
+        - make V=1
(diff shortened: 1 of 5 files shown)


dgibson:
Sorry I've lost this for years.  I'm not really used to the github PR workflow.

In any case, fdtdump is supposed to be a really minimal debugging tool, so I'm not going to apply this.

I would consider colorization for dtc -O dts output mode.

PR #12, opened 2017-12-06
Add limited read-only support for older (V2 and V3) device tree to libfdt.

nwhitehorn:
This can be useful in particular in the kernel when booting on systems with FDT-emitting firmware that is out of date. Releases of kexec-tools on ppc64 prior to the end of 2014 are notable examples of such.

The limitation to the read-only support is that the fdt_get_property() API is intrinsically incompatible with older versions of the device tree as a result of the 8-byte alignment of properties, so will return BADVERSION. The fdt_getprop() variant is, however, supported  on the older structures.

### DIFF
--- a/fdtget.c
+++ b/fdtget.c
@@ -140,7 +140,6 @@ static int show_data(struct display_info *disp, const char *data, int len)
  */
 static int list_properties(const void *blob, int node)
 {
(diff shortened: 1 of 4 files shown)


nwhitehorn:
Merged.

PR #13, opened 2017-12-11
Fix boundary check in fdt_strerror

balika011:
minus operator applied to unsigned type, result still unsigned -> the check always results in false and the function returns `"<unknown error>"`.
Also even if the check would pass it would do an out of bounds reading of the fdt_errtable table on fdt_strerror(FDT_ERRTABSIZE) function call.

### DIFF
--- a/libfdt/fdt_strerror.c
+++ b/libfdt/fdt_strerror.c
@@ -91,7 +91,7 @@ const char *fdt_strerror(int errval)
 		return "<valid offset/length>";
 	else if (errval == 0)
 		return "<no error>";
-	else if (errval > -FDT_ERRTABSIZE) {
+	else if (-errval < FDT_ERRTABSIZE) {
 		const char *s = fdt_errtable[-errval].str;
 
 		if (s)


dgibson:
On Mon, Dec 11, 2017 at 10:39:48AM +0000, Balázs Triszka wrote:
> minus operator applied to unsigned type, result still unsigned -> the check always results in false and the function returns "<unknown error>".
> Also even if the check would pass it would do an out of bounds reading of the fdt_errtable table on fdt_strerror(FDT_ERRTABSIZE) function call.
> You can view, comment on, or merge this pull request online at:

Nice catch.  However, I'll need a Linux-style signed-off-by line from
you in order to apply the fix.

Note that in general, the preferred workflow for dtc contributions is
to send patches as email to myself and the
<devicetree-compiler@vger.kernel.org> list, rather than using github
pull requests.

> 
>   https://github.com/dgibson/dtc/pull/13
> 
> -- Commit Summary --
> 
>   * Fix boundary check in fdt_strerror
> 
> -- File Changes --
> 
>     M libfdt/fdt_strerror.c (2)
> 
> -- Patch Links --
> 
> https://github.com/dgibson/dtc/pull/13.patch
> https://github.com/dgibson/dtc/pull/13.diff
> 

-- 
				| _way_ _around_!

PR #20, opened 2018-11-02
libfdt: return correct value if #size-cells property is not present

johnrclarke:
According to the device tree specification, the default value for
#size-cells is 1, but fdt_size_cells() was returning 2 if this property
was not present.

This patch also makes fdt_address_cells() and fdt_size_cells() conform
to the behaviour documented in libfdt.h. The defaults are only returned
if fdt_getprop() returns -FDT_ERR_NOTFOUND, otherwise the actual error
is returned.


johnrclarke:
Signed-off-by added.  Thanks.

Issue #23, opened 2018-11-20
Should use strcmp/strncmp instead of memcmp on arm.


xieyubo:
In libfdt code, seems we use memcmp to compare two strings in some places. It is not correct behavior on arm. Arm has memory alignment requirement on memcmp(). But for string, especially the string directly comes from dt string block, there is no guarantee that the address is aligned. memcmp() will cause an exception.

dgibson:
Um.. a memcmp() implementation which has alignment constraints is *utterly* broken.  I can't believe that's the case on all ARM systems - what specific environment are you using?

xieyubo:
compiler: gcc-arm-8.2-2018.08-x86_64-aarch64-elf which uses newlib
device: Raspberry Pi 3 B

xieyubo:
S:\src\pios>gcc-arm-8.2-2018.08-i686-mingw32-aarch64-elf\bin\aarch64-elf-gcc -v
Using built-in specs.
COLLECT_GCC=gcc-arm-8.2-2018.08-i686-mingw32-aarch64-elf\bin\aarch64-elf-gcc
COLLECT_LTO_WRAPPER=s:/src/pios/gcc-arm-8.2-2018.08-i686-mingw32-aarch64-elf/bin/../libexec/gcc/aarch64-elf/8.2.1/lto-wrapper.exe
Target: aarch64-elf
[...]

PR #24, opened 2018-11-20
Fix crash on arm due to memcmp on unaligned address.

xieyubo:
Issue: https://github.com/dgibson/dtc/issues/23


dgibson:
This looks deeply broken.  memcmp() should absolutely work on unaligned addresses, regardless of the architecture.  If it does not, something is wrong with your compiler or libc.

PR #29, opened 2019-08-20
libfdt: Remove semicolons after function definitions

davidbrazdil:
ISO C does not allow semicolons after function bodies. Remove them.


dgibson:
Change looks good, but it needs a proper commit message.

The rationale should be included  in the message, as well as a Linux-style Signed-off-by line.

PR #30, opened 2019-12-05
Add EXTRA_CXXFLAGS to CXXFLAGS.

marxin:


dgibson:
Your commit message has no description or rationale for this change.  You'll definitely need to add that before I'll merge.  You'll also need a Signed-off-by: line in the kernel style.

The actual change also doesn't immediately look right to me: you're adding all EXTRA_CXXFLAGS - not something that's usually used with dtc, but generally flags for the C++ compiler, to CPPFLAGS, which is flags for the C preprocessor.

marxin:
You are right, I don't need it, `EXTRA_CFLAGS` work fine foe me.
Thanks.

PR #34, opened 2020-05-08
libfdt: Make fdt_cells function accessible

mbgg:
For reading address-cells and size-cells, the libfdt only provides
functions which do not return in case the node does not provide the
property. For traversing the DT to find the parent node which provides
this property we will need to know that.

Make fdt_cells accessible from outside of libfdt so that we can detect
not present size- and address-cells properties in a node.

Signed-off-by: Matthias Brugger <mbrugger@suse.com>

### DIFF
--- a/libfdt/fdt_addresses.c
+++ b/libfdt/fdt_addresses.c
@@ -11,7 +11,7 @@
 
 #include "libfdt_internal.h"
 
-static int fdt_cells(const void *fdt, int nodeoffset, const char *name)
+int fdt_cells(const void *fdt, int nodeoffset, const char *name)
 {
 	const fdt32_t *c;
 	uint32_t val;
--- a/libfdt/libfdt.h
+++ b/libfdt/libfdt.h
@@ -1193,6 +1193,27 @@ int fdt_address_cells(const void *fdt, int nodeoffset);
  */
 int fdt_size_cells(const void *fdt, int nodeoffset);
 
+/**
+ * fdt_cells - retrive size of cells specified by parameter name.
+ * @fdt: pointer to the device tree blob
+ * @nodeoffset: offset of the node to find the address range size for
+ * @name: name of the property which size we want to read
+ *
+ * When the node has a valid property define in name, returns its value.
+ *
+ * returns:
+ *	0 <= n < FDT_MAX_NCELLS, on success
+ *      -FDT_ERR_NOTFOUND, if the node has no property specified in paramter
+		name
+ *      -FDT_ERR_BADNCELLS, if the node has a badly formatted or invalid
+ *		#size-cells property
+ *	-FDT_ERR_BADMAGIC,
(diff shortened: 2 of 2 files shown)


dgibson:
On Fri, May 08, 2020 at 03:43:05PM -0700, Matthias Brugger wrote:
> For reading address-cells and size-cells, the libfdt only provides
> functions which do not return in case the node does not provide the
> property. For traversing the DT to find the parent node which provides
> this property we will need to know that.

Uuh.. this sounds like your trees are doing the wrong thing.

#address-cells and #size-cells properties are *not* inherited from the
parent if the properties are missing.  Instead they revert to the
defaults (2 for #address-cells and 1 for #size-cells).  That's why
fdt_address_cells() and fdt_size_cells() are implemented the way they are.


> Make fdt_cells accessible from outside of libfdt so that we can detect
> not present size- and address-cells properties in a node.
> 
> Signed-off-by: Matthias Brugger &lt;mbrugger@suse.com&gt;
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/dgibson/dtc/pull/34
> 
> -- Commit Summary --
> 
>   * libfdt: Make fdt_cells function accessible
> 
> -- File Changes --
> 
>     M libfdt/fdt_addresses.c (2)
>     M libfdt/libfdt.h (21)
> 
> -- Patch Links --
> 
> https://github.com/dgibson/dtc/pull/34.patch
> https://github.com/dgibson/dtc/pull/34.diff
> 

-- 
				| _way_ _around_!

mbgg:
Well it is used in a child to find out which address-cells and size-cells it has.

"The #address-cells and #size-cells properties may be used in any device node that has children in the devicetree hierarchy and describes how child device nodes should be addressed." [1]

My understanding of this is that the children inherit the address-cells and size-cells from the parent. Otherwise we would need to put cell sizes in every single node.

[1] https://github.com/devicetree-org/devicetree-specification/blob/master/source/devicetree-basics.rst#address-cells-and-size-cells

dgibson:
On Mon, May 11, 2020 at 01:40:52AM -0700, Matthias Brugger wrote:
> Well it is used in a child to find out which address-cells and size-cells it has.
> 
> "The #address-cells and #size-cells properties may be used in any
> device node that has children in the devicetree hierarchy and
> describes how child device nodes should be addressed." [1]
> 
> My understanding of this is that the children inherit the
> address-cells and size-cells from the parent. Otherwise we would
> need to put cell sizes in every single node.

Well, depends what you mean by "inherit".  Basically #address-cells
and #size-cells describe the address space for the bus they're in.
Therefore the "reg" of a node is described by #address-cells and
#size-cells of the parent.  It's only and always the *immediate*
parent, though, never from any higher up the tree.  If the parent node
has no #address-cells and #size-cells, the relevant values are 2 and
1.

-- 
				| _way_ _around_!

PR #35, opened 2020-05-12
libfdt: Export overlay_apply_node() as fdt_overlay_apply_node()

marex:
This function is useful to merge a subset of DT into another DT, for
example if some prior-stage firmware passes a DT fragment to U-Boot
and U-Boot needs to merge it into its own DT. Export this function
to permit implementing such functionality.

### DIFF
--- a/libfdt/fdt_overlay.c
+++ b/libfdt/fdt_overlay.c
@@ -879,3 +879,8 @@ err:
 
 	return ret;
 }
+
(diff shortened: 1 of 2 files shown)


marex:
It seems this patch was not applied upstream, why ?

dgibson:
Sorry, I renamed the `master` branch to `main` and that seems to have auto-closed all the PRs against it.  Can you resubmit this PR against the `main` branch.

PR #36, opened 2020-05-21
start of /rename-node/ support

galak:
Signed-off-by: Kumar Gala <kumar.gala@linaro.org>

### DIFF
--- a/dtc-lexer.l
+++ b/dtc-lexer.l
@@ -137,6 +137,13 @@ static void PRINTF(1, 2) lexical_error(const char *fmt, ...);
 			return DT_DEL_NODE;
 		}
 
+<*>"/rename-node/"	{
+			DPRINT("Keyword: /rename-node/\n");
+			DPRINT("<PROPNODENAME>\n");
+			BEGIN(PROPNODENAME);
+			return DT_RENAME_NODE;
+		}
+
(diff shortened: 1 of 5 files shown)


galak:
@dgibson the lex/yacc for the subnode case here isn't correct, but not sure how to fix it.  Any help would be appreciated.

dgibson:
Uh... difficult to answer that without some more context on what you're trying to accomplish here.

galak:
The idea is there 2 use cases I had in mind:

1. There are a number of cases we've seen in which there is an IP block that can be configured to be either a UART, I2C, or SPI.  So being able to do something:

```
sercom5: sercom@4500000 {
   ...
};

/rename-node/ &sercom5 i2c@4500000;
```

2. There are cases in which the `reg` might get redefined for a node and than having `rename-node` allows us to fix the node name to match the `reg` property.

galak:
> Uh... difficult to answer that without some more context on what you're trying to accomplish here.

I was trying to mimic what existed for `/delete-node/`.

galak:
@dgibson curious why you closed this.  You never responded to my comments on what I was trying to accomplish.

dgibson:
Oh.. looks like renaming the master branch to main has auto-closed it :/.

PR #38, opened 2020-07-22
libfdt: fix signed comparion errors

vwadekar:
This patch fixes the signed comparison errors seen with the
"-Wsign-compare" compiler flag.

Signed-off-by: Varun Wadekar <vwadekar@nvidia.com>
Change-Id: I9b81e3f04b6420498f78b91b7ef031179c3c9959

### DIFF
--- a/libfdt/fdt.c
+++ b/libfdt/fdt.c
@@ -137,13 +137,13 @@ const void *fdt_offset_ptr(const void *fdt, int offset, unsigned int len)
 	unsigned absoffset = offset + fdt_off_dt_struct(fdt);
 
 	if (!can_assume(VALID_INPUT))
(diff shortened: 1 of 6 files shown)


Andre-ARM:
Hi Varun, I think the official contribution process is sending patches via the mailing list?
Anyway, @sjg20 posted a very similar patch back in January, which had some issues: https://www.spinics.net/lists/devicetree-compiler/msg02953.html
So I created a hopefully more sustainable version: https://www.spinics.net/lists/devicetree-compiler/msg03333.html

Can you check how this looks for you, then reply on the mailing list? Or at least check whether this fixes the issues for you?

I put a copy here: https://github.com/Andre-ARM/dtc/commits/signed-comparison

PR #40, opened 2020-08-11
Makefile: permit non-default DESTDIR and PREFIX

cxw42:
Make the assignments to DESTDIR and PREFIX conditional.  This permits installing to a particular directory using, e.g., `make install PREFIX=/usr/local`.

Thanks for considering this PR!


dgibson:
Uh.. as far as I can tell, this will have no effect.

Setting variables on the command line already overrides settings in the Makefile.  Using '?=' just means that we won't override settings earlier in the Makefile itself.

cxw42:
Almost 20 years with `make` and I didn't know that --- sorry for the bother, and thanks for letting me know!

PR #44, opened 2020-12-23
Endianness clarification: misleading types for load/store

fozog:
fdt32_ld() fdt64_ld() fdt32_st() and fdt64_st() were using host endian types
while they should indicate they load or store fdt32_t or fdt64_t values.

### DIFF
--- a/libfdt/libfdt.h
+++ b/libfdt/libfdt.h
@@ -130,7 +130,7 @@ uint32_t fdt_next_tag(const void *fdt, int offset, int *nextoffset);
  * to work even with unaligned pointers on platforms (such as ARMv5) that don't
  * like unaligned loads and stores.
  */
-static inline uint32_t fdt32_ld(const fdt32_t *p)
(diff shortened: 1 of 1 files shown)


dgibson:
This change is not correct.  The uint32_t (etc.) return values *correctly* indicate that the return value has been changed from the big-endian value in the dtb to host native endian.  Thats the whole point of these functions.

fozog:
> This change is not correct. The uint32_t (etc.) return values _correctly_ indicate that the return value has been changed from the big-endian value in the dtb to host native endian. Thats the whole point of these functions.

Actually, it translates to little-endian, not host native endian. That's what confused me. On a big endian system you have to call cpu_to_fdt64 on the returned value to use it.

Sorry for the invalid pull request that I made way to quick.... I commit to react more slowly next time ;-)

dgibson:
Uh, no, it translates to host native.  Look again.

PR #53, opened 2021-08-10
False effective head detection: fdt_totalsize(p) equal the len of dtb

### LINKED ISSUE #52: valid_header function may be have some problem.
https://github.com/dgibson/dtc/blob/ecfb438c07fa468a129e81c7d84c7c293c7b0150/fdtdump.c#L166
This function should be modified like this?
```
static bool valid_header(char *p, size_t len)
{
	if (len < sizeof(struct fdt_header) ||
	    fdt_magic(p) != FDT_MAGIC ||
	    fdt_version(p) > MAX_VERSION ||
	    fdt_last_comp_version(p) > MAX_VERSION ||
	    fdt_totalsize(p) > len ||
	    fdt_off_dt_struct(p) >= len ||
	    fdt_off_dt_strings(p) >= len)
		return 0;
	else
		return 1;
}
```

juiceRv:
file should be legitimate

Signed-off-by: juiceRv <juicemail@163.com>
https://github.com/dgibson/dtc/issues/52#issue-964778171

### DIFF
--- a/fdtdump.c
+++ b/fdtdump.c
@@ -169,7 +169,7 @@ static bool valid_header(char *p, size_t len)
 	    fdt_magic(p) != FDT_MAGIC ||
 	    fdt_version(p) > MAX_VERSION ||
 	    fdt_last_comp_version(p) > MAX_VERSION ||
-	    fdt_totalsize(p) >= len ||
+	    fdt_totalsize(p) > len ||
 	    fdt_off_dt_struct(p) >= len ||
 	    fdt_off_dt_strings(p) >= len)
 		return 0;


juiceRv:
#52

dgibson:
I think your change is correct, but your commit message needs to explain the problem and the solution more clearly.

In addition, Signed-off-by lines need to use your real name, not a pseudonym or handle.

juiceRv:
> I think your change is correct, but your commit message needs to explain the problem and the solution more clearly.
> 
> In addition, Signed-off-by lines need to use your real name, not a pseudonym or handle.


I am very honored to have your support, I will immediately correct my mistakes.

PR #54, opened 2021-10-31
dtc: introduce label relative path references

a3f:
Reference via label allows extending nodes with compile-time checking of
whether the node being extended exists. This is useful to catch
renamed/removed nodes after an update of the device trees to be extended.

In absence of labels in the original device trees, new style path
references can be used:

```
/* upstream device tree */
/ {
	leds: some-non-standard-led-controller-name {
		led-0 {
			default-state = "off";
		};
	};
};

/* downstream device tree */
&{/some-non-standard-led-controller-name/led-0} {
	default-state = "on";
};
```

This is a common theme within the barebox bootloader[0], which extends the
upstream (Linux) device trees in that manner. The downside is that,
especially for deep nodes, these references can get quite long and tend to
break often due to upstream rework (e.g. rename to adhere to bindings).

Often there is a label a level or two higher that could be used. This
patch allows combining both a label and a new style path reference to
get a compile-time-checked reference, which allows rewriting the
previous downstream device tree snippet to:

```
&leds/led-0 {
	default-state = "on";
};
```

This won't be broken when `/some-non-standard-led-controller-name` is
renamed or moved while keeping the label. And if `led-0` is renamed, we
will get the expected compile-time error.

Overlay support is skipped for now as they require special support: The
label and relative path parts need to be resolved at overlay apply-time,
not at compile-time.

[0]: https://www.barebox.org/doc/latest/devicetree/index.html

### DIFF
--- a/dtc-lexer.l
+++ b/dtc-lexer.l
@@ -200,7 +200,7 @@ static void PRINTF(1, 2) lexical_error(const char *fmt, ...);
 			return DT_LABEL_REF;
 		}
 
-<*>"&{/"{PATHCHAR}*\}	{	/* new-style path reference */
+<*>"&{"{PATHCHAR}*\}	{	/* new-style path reference */
 			yytext[yyleng-1] = '\0';
 			DPRINT("Ref: %s\n", yytext+2);
 			yylval.labelref = xstrdup(yytext+2);
--- a/dtc-parser.y
+++ b/dtc-parser.y
@@ -23,6 +23,12 @@ extern void yyerror(char const *s);
 
 extern struct dt_info *parser_output;
 extern bool treesource_error;
+
+static bool is_ref_relative(const char *ref)
+{
+	return ref[0] != '/' && strchr(&ref[1], '/');
+}
+
 %}
 
 %union {
@@ -169,6 +175,8 @@ devicetree:
 			 */
 			if (!($<flags>-1 & DTSF_PLUGIN))
 				ERROR(&@2, "Label or path %s not found", $1);
+			else if (is_ref_relative($1))
+				ERROR(&@2, "Label-relative reference %s not supported in plugin", $1);
 			$$ = add_orphan_node(
 					name_node(build_node(NULL, NULL, NULL),
 						  ""),
@@ -177,6 +185,9 @@ devicetree:
 	| devicetree DT_LABEL dt_ref nodedef
 		{
 			struct node *target = get_node_by_ref($1, $3);
+
+			if (($<flags>-1 & DTSF_PLUGIN) && is_ref_relative($3))
+				ERROR(&@2, "Label-relative reference %s not supported in plugin", $3);
 
 			if (target) {
 				add_label(&target->labels, $2);
@@ -193,6 +204,8 @@ devicetree:
 			 * so $-1 is what we want (plugindecl)
 			 */
 			if ($<flags>-1 & DTSF_PLUGIN) {
+				if (is_ref_relative($2))
+					ERROR(&@2, "Label-relative reference %s not supported in plugin", $2);
 				add_orphan_node($1, $3, $2);
 			} else {
 				struct node *target = get_node_by_ref($1, $2);
--- a/util.c
+++ b/util.c
@@ -33,6 +33,17 @@ char *xstrdup(const char *s)
 	return d;
 }
 
+char *xstrndup(const char *s, size_t n)
+{
+	size_t len = strnlen(s, n) + 1;
+	char *d = xmalloc(len);
+
+	memcpy(d, s, len - 1);
+	d[len - 1] = '\0';
+
+	return d;
+}
+
 int xavsprintf_append(char **strp, const char *fmt, va_list ap)
 {
 	int n, size = 0;	/* start with 128 bytes */
--- a/livetree.c
+++ b/livetree.c
@@ -581,12 +581,39 @@ struct node *get_node_by_phandle(struct node *tree, cell_t phandle)
 
 struct node *get_node_by_ref(struct node *tree, const char *ref)
 {
+	struct node *target = tree;
+	const char *label = NULL, *path = NULL;
+
 	if (streq(ref, "/"))
 		return tree;
-	else if (ref[0] == '/')
-		return get_node_by_path(tree, ref);
+
+	if (ref[0] == '/')
+		path = ref;
 	else
-		return get_node_by_label(tree, ref);
+		label = ref;
+
+	if (label) {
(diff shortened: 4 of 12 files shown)


a3f:
@dgibson Any feedback?

dgibson:
> @dgibson Any feedback?

Sorry I've taken so long.  I've been pretty busy with other things, and then I injured myself :/.

util.c
```diff
@@ -33,6 +33,17 @@ char *xstrdup(const char *s)
...
+char *xstrndup(const char *s, size_t n)
+{
+	size_t len = strnlen(s, n) + 1;
+	char *d = xmalloc(len);
+
+	memcpy(d, s, len - 1);
+	d[len - 1] = '\0';
+
+	return d;
+}
```
dgibson, line 45: Is there a compelling reason to do it this way, rather than just using `strndup()` and checking for a NULL return value?

dtc-lexer.l
```diff
@@ -207,6 +207,12 @@ static void PRINTF(1, 2) lexical_error(const char *fmt, ...);
 			return DT_PATH_REF;
 		}
 
+<*>\&{LABEL}\/{PATHCHAR}*	{	/* label-relative path reference */
+			DPRINT("Ref: %s\n", yytext+1);
+			yylval.labelref = xstrdup(yytext+1);
+			return DT_REL_REF;
+		}
+
```
dgibson, line 215: The basic idea seems good to me: this works with the many similarities between dtc labels and OF aliases (alias relative paths were traditionally allowed in OF).  Lexing this separately from the existing references makes me a little bit nervous though.

On the one hand, PATHCHAR is a sufficiently wide class, that I worry that lexing it without some sort of delimiters around it could lead to ambiguities in certain cases.  On the other hand, my general experience is that it's usually best to make the lexer to accept as much as you reasonably can for a particular token type, then filter it down in the parser.  That tends to lead to more useful error behaviour than having the lexer fall back to a different token type which then feeds nonsense into the parser.

At least for a first pass, I'd suggest making label-relative references a variant of absolute path references, so `&{label/then/subdirs}` rather than `&label/then/subdirs`.  I believe you can have the lexer accept both as the same token type, then later disambiguate them depending on whether they start with `/` or not.

dgibson (changes requested):

util.c
a3f: `strdup`/`strndup` are `_POSIX_C_SOURCE >= v200809L`. `xstrdup`, which already exists, isn't implemented in terms of `strdup`, so I followed suit for `xstrndup` as well.

---

Note: There is one `strdup` use in `tests/setprop_inplace.c`, everything else is xstrdup. I submitted #63 just now to address this.

dtc-lexer.l
a3f: Fair enough. I've just force-pushed syntax change to `&{label/then/subdirs}`.

a3f:
> Sorry I've taken so long. I've been pretty busy with other things, and then I injured myself :/.

Get well soon! Thanks for the feedback, I've just incorporated it.

util.c
dgibson: Ok, makes sense.

dtc-parser.y
```diff
@@ -169,6 +169,8 @@ devicetree:
 			 */
 			if (!($<flags>-1 & DTSF_PLUGIN))
 				ERROR(&@2, "Label or path %s not found", $1);
+			else if (strchr($1 + 1, '/'))
```
dgibson, line 172: I don't think this condition is quite right.  This determines that the reference is *not* a pure label, but it won't distinguish between an absolute path reference and a label-relative path reference.

dtc-parser.y
```diff
@@ -178,6 +180,9 @@ devicetree:
 		{
 			struct node *target = get_node_by_ref($1, $3);
 
+			if (($<flags>-1 & DTSF_PLUGIN) && strchr($3 + 1, '/'))
```
dgibson, line 183: Same issue here.

dgibson (changes requested):
Mostly looks good to me now, but there are a couple of remaining nits.

dtc-parser.y
a3f: Ah, yes. With the lexer changes, it's indeed not correct anymore. I defined a helper now to check it correctly and used it here.

dtc-lexer.l
```diff
@@ -200,7 +200,7 @@ static void PRINTF(1, 2) lexical_error(const char *fmt, ...);
 			return DT_LABEL_REF;
 		}
 
-<*>"&{/"{PATHCHAR}*\}	{	/* new-style path reference */
+<*>"&{"{PATHCHAR}*\}	{	/* new-style path reference */
```
dgibson, line 203: This is strictly broader than the `DT_LABEL_REF` pattern, so it should be possible to collapse both ref types into a single token type.

dtc-parser.y
```diff
@@ -23,6 +23,12 @@ extern void yyerror(char const *s);
 
 extern struct dt_info *parser_output;
 extern bool treesource_error;
+
+static bool is_label_relative(const char *name)
```
dgibson, line 27: Rename to `is_ref_relative` since if it returns true the given string is *not* (just) a label.  Rename `name` to `ref` would also be clearer.

dgibson (changes requested):
Sorry it's taken me so long to look at this. The main part is looking good, but there are still some cleanups I'd prefer.

dtc-lexer.l
a3f: Can't follow you here. `DT_LABEL_REF` is `<*>\&{LABEL}` (unescaped `{}`, e.g. `&label`), while `DT_PATH_REF` is `<*>"&{"{PATHCHAR}*\}` (escaped `{}`, e.g. `&{label/}`).

So no pattern is strictly broader than the other?

dtc-parser.y
a3f: Done.

dtc-lexer.l
dgibson: Oh, sorry, completely missed the `{}`.  Disregard.

dgibson (approved):

dgibson:
Merged, thanks!

a3f:
Thank you! :)

a3f:
Just fixed a regression that would've been avoidable with a relative path reference: https://lore.barebox.org/barebox/20220811192612.3021687-1-ahmad@a3f.at/T/#u

Will there perhaps be a new dtc release soon? :-)

PR #55, opened 2021-10-31
CI: Cirrus: bump used FreeBSD from 12.1 to 13.0

a3f:
CI freebsd_12 job currently fails to build PRs, because of:
```
ld-elf.so.1: /usr/local/bin/bison: Undefined symbol "fread_unlocked@FBSD_1.6"
```

According to FreeBSD issue tracker[1], the proper solution is to upgrade to a
supported release, so do that for our CI.

[1]: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=253452

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,9 +1,9 @@
 env:
   CIRRUS_CLONE_DEPTH: 1
 
-freebsd_12_task:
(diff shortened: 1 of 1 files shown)


dgibson:
Merged, thanks.

PR #57, opened 2021-11-09
libfdt: fix an incorrect integer promotion

parport0:
UINT32_MAX is an integer of type unsigned int. UINT32_MAX + 1 overflows unless explicitly computed as unsigned long long. This led to some invalid addresses being treated as valid.

Specify that 1 is an unsigned long long integer.

### DIFF
--- a/libfdt/fdt_addresses.c
+++ b/libfdt/fdt_addresses.c
@@ -73,7 +73,7 @@ int fdt_appendprop_addrrange(void *fdt, int parent, int nodeoffset,
 	/* check validity of address */
 	prop = data;
 	if (addr_cells == 1) {
-		if ((addr > UINT32_MAX) || ((UINT32_MAX + 1 - addr) < size))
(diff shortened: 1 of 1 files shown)


libfdt/fdt_addresses.c
```diff
@@ -73,7 +73,7 @@ int fdt_appendprop_addrrange(void *fdt, int parent, int nodeoffset,
 	/* check validity of address */
 	prop = data;
 	if (addr_cells == 1) {
-		if ((addr > UINT32_MAX) || ((UINT32_MAX + 1 - addr) < size))
+		if ((addr > UINT32_MAX) || ((UINT32_MAX + 1ULL - addr) < size))
```
dgibson, line 76: Nice catch.  However, a small detail, I'd prefer to see an explicit case of UINT32_MAX to (uint64_t), which should enforce the same promotion.  `1ULL` will probably have the same type as `uint64_t` and therefore as `addr` and `size`, but it's nice to be explicit.

parport0: Alright.

PR #61, opened 2021-11-11
pylibfdt: fix with Python 3.10

rossburton:
Since Python 2.5 the argument parsing functions when parsing expressions
such as s# (string plus length) expect the length to be an int or a
ssize_t, depending on whether PY_SSIZE_T_CLEAN is defined or not.

Python 3.8 deprecated the use of int, and with Python 3.10 this symbol
must be defined and ssize_t used[1].

Define the magic symbol when building the extension, and cast the ints
from the libfdt API to ssize_t as appropriate.

[1] https://docs.python.org/3.10/whatsnew/3.10.html#id2

Signed-off-by: Ross Burton <ross.burton@arm.com>

### DIFF
--- a/pylibfdt/libfdt.i
+++ b/pylibfdt/libfdt.i
@@ -1044,9 +1044,9 @@ typedef uint32_t fdt32_t;
 		$result = Py_None;
 	else
         %#if PY_VERSION_HEX >= 0x03000000
-            $result = Py_BuildValue("y#", $1, *arg4);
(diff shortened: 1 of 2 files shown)


dgibson:
For future reference, you don't need to send a github PR *and* a direct email.  Either one is fine.  In any case I've applied this with a manual correction from the email version, so I'm closing this.

rossburton:
Wasn't sure what the 'preferred' method is.  If both are good then I'll stick with PRs.

PR #63, opened 2021-12-09
tests: setprop_inplace: use xstrdup instead of unchecked strdup

a3f:
This is the only `strdup` instance we have, all others are `xstrdup`. As
`strdup` is `_POSIX_C_SOURCE >= v200809L`, which we don't require and we
don't check `strdup` error return here, switch to `xstrdup` instead. This
aligns the test with others that call `x`funcs, mainly `xmalloc()`.


dgibson:
Sorry, I somehow missed this one.  Merged now.

Issue #62, opened 2021-11-21
improve annotation


bmx666:
Hi, I'm bringing back to life ["Device Tree Visualizer"](https://github.com/bmx666/dtv-demo)
This project is helpful for our team for analyze complex projects where multiple DTS included and overlayed.

I made a few WORKAROUNDs inside [my fork](https://github.com/bmx666/dtc/commits/main).
But I'm not expert in DT compiler and it's hard to imagine whole picture "how dtc works".

Is it possible to generate compatible DTS which highlight "deleted nodes/props" but when you pass special flag?

dgibson:
I'm afraid I'm really not sure what you're asking form.

bmx666:
@dgibson for example, command to generate full annotated final DTS file

```
cpp -nostdinc -undef -D__DTS__ \
    -x assembler-with-cpp \
    -I ~/linux/include/ -I ~/linux/arch/arm/boot/dts -I ~/linux/scripts/dtc/include-prefixes/ \
    ~/linux/arch/arm64/boot/dts/allwinner/sun50i-h5-bananapi-m2-plus.dts | \
    | dtc -I dts -O dts -f -s -T -T -o - > annotated.dts
```
 
### Problem 1
dtc put in comment only the latest merged source file for property, but left all source files for nodes. Example:

#### A.dts
```
my_node {
    test_A;
};
```

#### B.dts
```
#include "A.dts"
my_node {
    test;
};
```

#### C.dts
```
#include "B.dts"
my_node {
    test;
};
```

```
my_node { /* A.dts:0-0:0-0, B.dts:0-0:0-0, C.dts:0-0:0-0 */
    test_A;  /* A.dts:0-0:0-0 */
    test;  /* C.dts:0-0:0-0 */
};  /* A.dts:0-0:0-0, B.dts:0-0:0-0, C.dts:0-0:0-0 */
```

But I expected "B.dts" for "test" property too, like this:

```
my_node { /* A.dts:0-0:0-0, B.dts:0-0:0-0, C.dts:0-0:0-0 */
    test_A;  /* A.dts:0-0:0-0 */
    test;  /* B.dts:0-0:0-0, C.dts:0-0:0-0 */
};  /* A.dts:0-0:0-0, B.dts:0-0:0-0, C.dts:0-0:0-0 */
```

### Problem 2
In current version not possible to trace where "/delete-node/" or "/delete-property/" was called. For example:

#### A.dts
```
my_node {
    test_A;
};
```

[...]

PR #67, opened 2022-01-05
annotations: extend source position information for properties

bmx666:
merge_nodes, defined in livetree.c, uses srcpos_extend only for nodes.

### DIFF
--- a/livetree.c
+++ b/livetree.c
@@ -174,8 +174,8 @@ struct node *merge_nodes(struct node *old_node, struct node *new_node)
 
 				old_prop->val = new_prop->val;
 				old_prop->deleted = 0;
-				free(old_prop->srcpos);
-				old_prop->srcpos = new_prop->srcpos;
+				old_prop->srcpos = srcpos_extend(
+					old_prop->srcpos, new_prop->srcpos);
 				free(new_prop);
 				new_prop = NULL;
 				break;


dgibson:
> merge_nodes, defined in livetree.c, uses srcpos_extend only for nodes.

Sorry it's taken me forever to look at this.  However, this behaviour is intentional.   Because the new property value entirely overwrites the old property value, I didn't think it made sense to include the now obsolete location in the annotations.

bmx666:
Sorry @dgibson but then for nodes it's look a little bit strange, because you can remove whole node and re-create same node in different dts, BUT obsolete location will be left.

dgibson:
> Sorry @dgibson but then for nodes it's look a little bit strange, because you can remove whole node and re-create same node in different dts, BUT obsolete location will be left.

That's a fair point.  We should probably remove the old node srcpos information when we a `/delete-node/` directive is processed.

bmx666:
@dgibson I think after that a fair point, you understood I wanted to trace history path of property/node/label from first dts up to the latest when I asked about #62

dgibson:
@bmx666, obviously I haven't looked at this in ages.  Sorry.

If you still want to pursue this, can you make a case for the behaviour change you'd like to see, then we can go from there.

bmx666:
Hi @dgibson 
Sync with main branch without any conflicts.

PR #72, opened 2022-09-15
libdtc: remove duplicate judgments

jiaxianhua:
There is no need to check the VALID_DTB repeatedly, and can be combined into one if statement.

Signed-off-by: Jia Xianhua <jiaxianhua@gmail.com>


dgibson:
Merged, thanks.

PR #79, opened 2023-02-21
Some build fixes

elmarco:
Those few commits allow dtc to be built by qemu with a simple dtc.wrap file such as:

```
[wrap-git]
url = https://github.com/dgibson/dtc.git
revision = tag-or-sha1afterfix

[provide]
fdt = libfdt_dep
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -14,7 +14,8 @@ add_project_arguments(
     '-Wstrict-prototypes',
     '-Wmissing-prototypes',
     '-Wredundant-decls',
-    '-Wshadow'
+    '-Wshadow',
+    '-Wsuggest-attribute=format',
   ]),
   language: 'c'
 )
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,8 @@ ASSUME_MASK ?= 0
 
 CPPFLAGS = -I libfdt -I . -DFDT_ASSUME_MASK=$(ASSUME_MASK)
 WARNINGS = -Wall -Wpointer-arith -Wcast-qual -Wnested-externs -Wsign-compare \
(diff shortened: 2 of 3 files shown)


elmarco:
@dgibson any idea why?:
``
 fdtget-runtest.sh "MyBoardName\0MyBoardFamilyName\0" -tr label01.dts.fdtget.test.dtb / compatible:	FAIL Results differ from expected
``

dgibson:
I've cherry-picked patches 2 & 3.  I'm not really happy about patch 1, see review.

meson.build
```diff
@@ -14,7 +14,9 @@ add_project_arguments(
     '-Wstrict-prototypes',
     '-Wmissing-prototypes',
     '-Wredundant-decls',
-    '-Wshadow'
+    '-Wshadow',
+    '-Wno-error=discarded-qualifiers',
```
dgibson, line 18: I would indeed prefer to fix the warnings rather than just suppress them.

But... when I tried enabling `-Wdiscarded-qualifiers` on my system (Fedora 37, gcc-12.2.1-4.fc37.x86_64), I didn't get any warnings.  What are you seeing with this one?

meson.build
```diff
@@ -14,7 +14,9 @@ add_project_arguments(
...
+    '-Wno-error=suggest-attribute=format',
```
dgibson, line 19: I've added `-Wsuggest-attribute=format` to the default build flags (make & meson), and fixed the one warning that created.

meson.build
elmarco: @dgibson actually it triggers when using `-Wwrite-strings`

PR #82, opened 2023-02-27
Fix -Wwrite-strings

elmarco:
Those changes break API/ABI (some user char * are no longer free()), but they should also fix incorrect or unwanted free() calls.

### DIFF
--- a/checks.c
+++ b/checks.c
@@ -31,7 +31,7 @@ typedef void (*check_fn)(struct check *c, struct dt_info *dti, struct node *node
 struct check {
 	const char *name;
 	check_fn fn;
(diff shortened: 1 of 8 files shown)


dgibson:
So, I think this kind of just papers over the problem.

The basic problem here is that sometimes the `name` fields in properties and nodes are allocated (so should be `free()ed`) and other times are static strings.  I believe the original intention was that they always be allocated (so static strings should be `strdup()`ed before insertion), but that's a really easy constraint to miss, so we have.

I think that's what we really need to address.

elmarco:
See #86

PR #85, opened 2023-02-28
checks: correct I2C 10-bit address check

mranostay:
Currently if there is a valid 10-bit address the following warning is always displayed due to the 7-bit check failing due to reg > 0x7f

"I2C address must be less than 7-bits, got "0x800000a6". Set I2C_TEN_BIT_ADDRESS for 10 bit addresses or fix the property"

Fix this issue by checking if a 10-bit address is expected, and is valid in separate if statements.

Fixes: 8259d59f ("checks: Improve i2c reg property checking")


dgibson:
Applied, thanks.

PR #88, opened 2023-03-08
pylibfdt: fix compatibility with python2

jpalus:
c001fc01a43e ("pylibfdt: fix swig build in install") introduced python3-only super() invocation that fails with python2

  File "setup.py", line 48, in run
    return super().run()
TypeError: super() takes at least 1 argument (0 given)

while setup.py still claims to be compatible with both python versions:

>  While Python 3 is the default, it's also possible to invoke
>  this setup.py script with Python 2.

use legacy syntax that works in all python versions

Fixes: c001fc01a43e ("pylibfdt: fix swig build in install")

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ libfdt_module = Extension(
 class build_py(_build_py):
     def run(self):
         self.run_command("build_ext")
-        return super().run()
+        return _build_py.run(self)
 
 setup(
     name='libfdt',


setup.py
```diff
@@ -45,7 +45,7 @@ def get_top_builddir():
 class build_py(_build_py):
     def run(self):
         self.run_command("build_ext")
-        return super().run()
+        return _build_py.run(self)
```
dgibson, line 48: I thought in modern python2 is was preferred to use super(), just with explicit arguments, rather than invoking as an unbound method.

jpalus: Unfortunately it all depends on `build_py` class from setuptools referenced in `setup.py` as `_build_py` which is a super class. If it's so called "new-style" class `super()` can be used, but if it's "old-style" class `super()` cannot be used. The difference between "new-style" and "old-style" class is inheritance from `object` (for python 2 only, in python 3 all classes are "new-style"). Example:
```py
class Old:
    def __init__(self):
        pass

class New(object):
    def __init__(self):
        pass

class SubOld(Old):
    def __init__(self):
        super(Old, self).__init__()

class SubNew(New):
    def __init__(self):
        super(New, self).__init__()
```
Attempt to create instance of `SubNew` succeeds but trying to create object of `SubOld` fails:
```
Traceback (most recent call last):
  File "test.py", line 17, in <module>
    SubOld()
  File "test.py", line 11, in __init__
    super(Old, self).__init__()
TypeError: super() argument 1 must be type, not classobj
```

Unfortunately  `_build_py` is an "old-style" class at least in my env. So in the end decided to use legacy syntax that works for both python 2 and 3 and for both "old-style" and "new-style" classes.

dgibson: Oof, I see.  I'd forgotten that old-style classes need that - and I'm bit horrified to discover that setuptools is still relying on an old-style class.

My misgiving about this is that doing it this way means we're not using `super()` for Python3 either.  `super()`'s behaviour is slightly different from just directly calling the superclass's method.  As far as I know the difference won't matter unless there's multiple inheritance going on, which I wouldn't expect to be the case here.  Nonetheless I don't love the idea of doing "the wrong thing" for Python3 for the sake of compatibility with not just Python2, but a pretty old concept in Python2.

jpalus:
No longer relevant since 3501d37.

PR #89, opened 2023-03-09
allow using labels with fdtget/fdtput

Villemoes:
It can be convenient and more readable to use `&label` syntax with fdtget and fdtput, rather than having to know the full path. This small series adds that support by hooking into fdt_path_offset_namelen() in the same place where we already allow the first component to be an alias. Of course this only works for blobs that have been built with -@, but it shouldn't affect existing use cases.

### DIFF
--- a/libfdt/fdt_ro.c
+++ b/libfdt/fdt_ro.c
@@ -258,14 +258,18 @@ int fdt_path_offset_namelen(const void *fdt, const char *path, int namelen)
 	if (!can_assume(VALID_INPUT) && namelen <= 0)
 		return -FDT_ERR_BADPATH;
 
-	/* see if we have an alias */
+	/* see if we have an alias or symbol */
 	if (*path != '/') {
 		const char *q = memchr(path, '/', end - p);
 
 		if (!q)
 			q = end;
 
-		p = fdt_get_alias_namelen(fdt, p, q - p);
+		if (*p == '&')
+			p = fdt_get_symbol_namelen(fdt, p + 1, q - p - 1);
+		else
+			p = fdt_get_alias_namelen(fdt, p, q - p);
+
 		if (!p)
 			return -FDT_ERR_BADPATH;
 		offset = fdt_path_offset(fdt, p);
--- a/libfdt/libfdt.h
+++ b/libfdt/libfdt.h
@@ -525,9 +525,10 @@ int fdt_path_offset_namelen(const void *fdt, const char *path, int namelen);
  * address).
  *
  * If the path is not absolute (i.e. does not begin with '/'), the
- * first component is treated as an alias.  That is, the property by
- * that name is looked up in the /aliases node, and the value of that
- * property used in place of that first component.
+ * first component is treated as an alias (or, if it begins with '&',
+ * the rest is treated as a symbol).  That is, the property by that
+ * name is looked up in the /aliases (or /__symbols__) node, and the
+ * value of that property used in place of that first component.
  *
  * For example, for this small fragment
  *
@@ -547,12 +548,17 @@ int fdt_path_offset_namelen(const void *fdt, const char *path, int namelen);
  *
  *   /soc@0/i2c@30a40000/eeprom@52
  *   i2c2/eeprom@52
+ *   &foo/eeprom@52
+ *   &bar
+ *
+ * The latter two only work if the device tree blob has been compiled
+ * with the -@ dtc option.
  *
  * returns:
  *	structure block offset of the node with the requested path (>=0), on
  *		success
  *	-FDT_ERR_BADPATH, given path does not begin with '/' and the first
- *		component is not a valid alias
+ *		component is not a valid alias or symbol
  *	-FDT_ERR_NOTFOUND, if the requested node does not exist
  *      -FDT_ERR_BADMAGIC,
  *	-FDT_ERR_BADVERSION,
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -885,9 +885,10 @@ dtbs_equal_tests () {
 }
 
 fdtget_tests () {
+    at="$1"
     dts=label01.dts
     dtb=$dts.fdtget.test.dtb
-    run_dtc_test -O dtb -o $dtb "$SRCDIR/$dts"
+    run_dtc_test $at -O dtb -o $dtb "$SRCDIR/$dts"
 
     # run_fdtget_test <expected-result> [<flags>] <file> <node> <property>
     run_fdtget_test "MyBoardName" $dtb / model
@@ -906,6 +907,11 @@ fdtget_tests () {
     # Here the property size is not a multiple of 4 bytes, so it should fail
     run_wrap_error_test $DTGET -tlx $dtb /randomnode mixed
     run_fdtget_test "6162 6300 1234 0 a 0 b 0 c" -thx $dtb /randomnode mixed
+    if [ -n "$at" ]; then
+        run_fdtget_test "6162 6300 1234 0 a 0 b 0 c" -thx $dtb "&node" mixed
+    else
+        run_wrap_error_test $DTGET  -thx $dtb "&node" mixed
+    fi
     run_fdtget_test "61 62 63 0 12 34 0 0 0 a 0 0 0 b 0 0 0 c" \
 	-thhx $dtb /randomnode mixed
     run_wrap_error_test $DTGET -ts $dtb /randomnode doctor-who
@@ -921,12 +927,13 @@ fdtget_tests () {
 }
 
 fdtput_tests () {
+    at="$1"
     dts=label01.dts
     dtb=$dts.fdtput.test.dtb
     text="$SRCDIR/lorem.txt"
 
     # Allow just enough space for $text
-    run_dtc_test -O dtb -p $($STATSZ $text) -o $dtb "$SRCDIR/$dts"
+    run_dtc_test $at -O dtb -p $($STATSZ $text) -o $dtb "$SRCDIR/$dts"
 
     # run_fdtput_test <expected-result> <file> <node> <property> <flags> <value>
     run_fdtput_test "a_model" $dtb / model -ts "a_model"
@@ -940,12 +947,19 @@ fdtput_tests () {
     run_fdtput_test "a0b0c0d deeaae ef000000" $dtb /randomnode blob \
 	-tx "a0b0c0d deeaae ef000000"
     run_fdtput_test "$(cat $text)" $dtb /randomnode blob -ts "$(cat $text)"
+    if [ -n "$at" ]; then
+        run_fdtput_test "y" $dtb '&node' x -ts "y"
+        run_fdtput_test "z" $dtb '&node/child' x -ts "z"
+    else
+        run_wrap_error_test $DTPUT $dtb '&node' x -ts "y"
+        run_wrap_error_test $DTPUT $dtb '&node/child' x -ts "z"
+    fi
 
     # Test expansion of the blob when insufficient room for property
     run_fdtput_test "$(cat $text $text)" $dtb /randomnode blob -ts "$(cat $text $text)"
 
     # Start again with a fresh dtb
-    run_dtc_test -O dtb -p $($STATSZ $text) -o $dtb "$SRCDIR/$dts"
+    run_dtc_test $at -O dtb -p $($STATSZ $text) -o $dtb "$SRCDIR/$dts"
 
     # Node creation
     run_wrap_error_test $DTPUT $dtb -c /baldrick sod
@@ -973,7 +987,7 @@ fdtput_tests () {
     run_wrap_test $DTPUT $dtb -cp /chosen/son
 
     # Start again with a fresh dtb
-    run_dtc_test -O dtb -p $($STATSZ $text) -o $dtb "$SRCDIR/$dts"
+    run_dtc_test $at -O dtb -p $($STATSZ $text) -o $dtb "$SRCDIR/$dts"
 
     # Node delete
     run_wrap_test $DTPUT $dtb -c /chosen/node1 /chosen/node2 /chosen/node3
@@ -1151,10 +1165,12 @@ for set in $TESTSETS; do
 	    dtbs_equal_tests
 	    ;;
 	"fdtget")
-	    fdtget_tests
+	    fdtget_tests ""
+	    fdtget_tests "-@"
 	    ;;
 	"fdtput")
-	    fdtput_tests
+	    fdtput_tests ""
+	    fdtput_tests "-@"
 	    ;;
 	"fdtdump")
 	    fdtdump_tests


Villemoes:
Any comments?

The CI errors don't seem to be related to these changes, so I don't know if I'm supposed to do anything about them.

libfdt/fdt_ro.c
```diff
@@ -466,6 +466,17 @@ const void *fdt_getprop_namelen(const void *fdt, int nodeoffset,
 	return prop->data;
```
dgibson, line 479: Hmm.. I think I need a bit more convincing that this is sufficiently widely useful to be worth adding.  It is, after all, just a wrapper over two consecutive calls.

libfdt/fdt_ro.c
```diff
@@ -247,6 +247,12 @@ int fdt_subnode_offset(const void *fdt, int parentoffset,
 	return fdt_subnode_offset_namelen(fdt, parentoffset, name, strlen(name));
 }
 
+static const char *fdt_get_symbol_namelen(const void *fdt,
```
dgibson, line 250: This seems sufficiently useful that it would be worth making public.

libfdt/fdt_ro.c
```diff
@@ -255,14 +261,18 @@ int fdt_path_offset_namelen(const void *fdt, const char *path, int namelen)
...
-	/* see if we have an alias */
+	/* see if we have an alias or symbol */
 	if (*path != '/') {
 		const char *q = memchr(path, '/', end - p);
 
 		if (!q)
 			q = end;
 
-		p = fdt_get_alias_namelen(fdt, p, q - p);
+		if (*p == '&')
```
dgibson, line 271: I'd like to see the new syntax described in the header comment for fdt_path_offset_namelen().

tests/run_tests.sh
```diff
@@ -859,7 +859,7 @@ dtbs_equal_tests () {
 fdtget_tests () {
     dts=label01.dts
     dtb=$dts.fdtget.test.dtb
-    run_dtc_test -O dtb -o $dtb "$SRCDIR/$dts"
+    run_dtc_test -@ -O dtb -o $dtb "$SRCDIR/$dts"
```
dgibson, line 862: Obviously you need to add `-@` to test the new functionality.  However, this means you're no longer testing the old functionality on a tree that's not compiled with `-@`.  I'd prefer to extend this so that both the `-@` and not `-@` cases are exercised.

dgibson:
Sorry, I'm finding it very difficult these days to find even minimal time to maintain dtc.

The concept for these patches looks good, I've put in a review for some of the details.

libfdt/fdt_ro.c
Villemoes: OK. Will refactor a bit so it gets added by itself then, and then used in fdt_path_offset_namelen().

tests/run_tests.sh
Villemoes: Good point. I'll see if I can figure out an easy way to do these tests both ways, and then either omit the new tests when not built with -@, or turn them into expected errors.

Villemoes:
Branch updated. I think this should address the comments, but I may have been too verbose in the fdt_path_offset documentation update.

libfdt/fdt_ro.c
```diff
@@ -269,6 +269,9 @@ int fdt_path_offset_namelen(const void *fdt, const char *path, int namelen)
 
 		if (!p)
 			return -FDT_ERR_BADPATH;
+		if (!can_assume(VALID_INPUT) && *p != '/')
+			return -FDT_ERR_BADVALUE;
+
```
ukleinek, line 274: Would it make sense to write:

    /aliases {
        self = "&somelabel";
    }
? With the previous commit it would work I think. I think I would have picked `-EFDT_ERR_BADPATH` as error code.

libfdt/fdt_ro.c
```diff
@@ -522,6 +522,18 @@ uint32_t fdt_get_phandle(const void *fdt, int nodeoffset)
 	return fdt32_ld_(php);
 }
 
+static const void *fdt_path_getprop_namelen(const void *fdt, const char *node,
+					    const char *name, int namelen,
+					    int *lenp)
```
ukleinek, line 534: I think the parameter naming is not optimal. `node` isn't a `struct node` but the path name of the node to lookup `name` in. Maybe use `s/node/nodename/` and `s/name/propname/`?

libfdt/libfdt.h
```diff
@@ -525,9 +525,10 @@ int fdt_path_offset_namelen(const void *fdt, const char *path, int namelen);
...
  *
  * If the path is not absolute (i.e. does not begin with '/'), the
- * first component is treated as an alias.  That is, the property by
- * that name is looked up in the /aliases node, and the value of that
- * property used in place of that first component.
+ * first component is treated as an alias (or, if it begins with '&',
+ * the rest is treated as a symbol).  That is, the property by that
+ * name is looked up in the /aliases (or /__symbols__) node, and the
+ * value of that property used in place of that first component.
  *
```
ukleinek, line 532: Something that is missing here is that the components in `path` don't need to specify the unit address ...

libfdt/libfdt.h
```diff
@@ -547,12 +548,17 @@ int fdt_path_offset_namelen(const void *fdt, const char *path, int namelen);
...
  *   /soc@0/i2c@30a40000/eeprom@52
  *   i2c2/eeprom@52
+ *   &foo/eeprom@52
+ *   &bar
```
ukleinek, line 552: ... so (as `/soc@0` is the only `soc@????` in `/` and `/soc@0/i2c@30a40000` is the only `i2c@???` in `/soc@0`, and `eeprom@52` is also unique,

    /soc/i2c/eeprom

is another path to that node.

I'm unsure if this is a bug of a feature though.

ukleinek (changes requested):
In general I like this series. Only a few little comments in the diffs.

The PR I added a few hours ago could benefit from the functions that are introduced here. I hesitate to base my work on top of this one, but once this PR is in shape and goes in I can rework my commits..

libfdt/fdt_ro.c
Villemoes: > Would it make sense to write:
> 
> ```
> /aliases {
>     self = "&somelabel";
> }
> ```
> 
> ? With the previous commit it would work I think. 

I can't see the use case. For this to resolve correctly, the dtb would have to have been built with `-@`, and then the `somelabel` property would still contain the absolute path, so there's no way this could even make the dtb smaller. So no, I would consider such a .dts file malformed, and I don't see any reason we should allow anything but absolute paths on the RHS of alias and symbol properties.

> I think I would have picked `-EFDT_ERR_BADPATH` as error code.

I believe my rationale in the commit message makes sense, and I certainly don't want to overload FDT_ERR_BADPATH more when there's such a fitting new error code to use.

libfdt/fdt_ro.c
Villemoes: Ack, grepping clearly shows that we only have `const char *nodename` and `const char *node_name`. And `propname` is also not unheard of, and makes sense here where there's also a nodename.

libfdt/libfdt.h
Villemoes: Huh? Try looking at the commit (or the "improve documentation for fdt_path_offset()") with `-U10`. Then you'll see that that is right there in the context:
```
--- a/libfdt/libfdt.h
+++ b/libfdt/libfdt.h
@@ -517,24 +517,49 @@ int fdt_path_offset_namelen(const void *fdt, const char *path, int namelen);
  * @fdt: pointer to the device tree blob
  * @path: full path of the node to locate
  *
  * fdt_path_offset() finds a node of a given path in the device tree.
  * Each path component may omit the unit address portion, but the
  * results of this are undefined if any such path component is
  * ambiguous (that is if there are multiple nodes at the relevant
  * level matching the given component, differentiated only by unit
  * address).
  *
+ * If the path is not absolute (i.e. does not begin with '/'), the
+ * first component is treated as an alias.  That is, the property by
+ * that name is looked up in the /aliases node, and the value of that
+ * property used in place of that first component.
```

Unless I'm misunderstanding what you were trying to say, I don't think there's anything to fix.

libfdt/libfdt.h
Villemoes: Well, in that very minimal example, yes, and presumably soc@0 is always the only soc node, but in real life, there'd very likely be several `i2c@...` nodes, and I don't see the point of confusing this example with the subtleties from that ambiguity, which has nothing to do with the feature I'm implementing.

libfdt/fdt_ro.c
ukleinek: > For this to resolve correctly, the dtb would have to have been built with `-@`, and then the `somelabel` property would still contain the absolute path, so there's no way this could even make the dtb smaller. So no, I would consider such a .dts file malformed, and I don't see any reason we should allow anything but absolute paths on the RHS of alias and symbol properties.

It can be smaller; with the following dts in test.dts

```dts
/dts-v1/;

/ {
	aliases {
		sl1 = "&somelabel";
		sl2 = &somelabel;
	};

	somelabel: somenodewithalongname {
		someproperty = "somevalue";
	};
};
```

and

```diff
diff --git a/libfdt/fdt_ro.c b/libfdt/fdt_ro.c
index d30c757180b7..c06c5d24d959 100644
--- a/libfdt/fdt_ro.c
+++ b/libfdt/fdt_ro.c
@@ -269,8 +269,10 @@ int fdt_path_offset_namelen(const void *fdt, const char *path, int namelen)
 
                if (!p)
                        return -FDT_ERR_BADPATH;
+#if 0
                if (!can_assume(VALID_INPUT) && *p != '/')
                        return -FDT_ERR_BADVALUE;
+#endif
 
                offset = fdt_path_offset(fdt, p);
 
```

```shell
$ dtc -@ -I dts -O dtb test.dts > test.dtb
test.dts:5.3-22: Warning (alias_paths): /aliases:sl1: aliases property is not a valid node (&somelabel)
$ fdtget test.dtb sl1 someproperty
somevalue
```

So it works fine and the sl1 alias is shorter than the equivalent sl2 one.

Villemoes: > > For this to resolve correctly, the dtb would have to have been built with `-@`, and then the `somelabel` property would still contain the absolute path, so there's no way this could even make the dtb smaller. So no, I would consider such a .dts file malformed, and I don't see any reason we should allow anything but absolute paths on the RHS of alias and symbol properties.
> 
> It can be smaller; with the following dts in test.dts
> 

Yes, of course, if the base assumption is that the dts will be built with -@ regardless, so the absolute path will be listed under the __symbols__ node, of course the string `&somelabel` will be shorter than the having the value of `sl1` be another copy of that absolute path.

However, that is entirely besides the point. I still completely fail to see the actual use case, or why we should allow aliases or symbols to resolve to anything but an absolute path, and stand by my "I would consider such a dts malformed". And current dtc obviously agrees with me, per the warning output above.

ukleinek: > > I think I would have picked `-EFDT_ERR_BADPATH` as error code.
> 
> I believe my rationale in the commit message makes sense, and I certainly don't want to overload FDT_ERR_BADPATH more when there's such a fitting new error code to use.

Hm, to sum up, with

```
/dts-v1/;

/ {
	aliases {
		sa1 = &somelabel;
		sa2 = "/nonexisting";
		sa3 = "invalid";
	};

	somelabel: somenodewithalongname {
		someproperty = "somevalue";
	};
};
```

we get for `fdtget test.dtb sa$i someproperty` respectively:
 * `i = 1`, success: `"somevalue"`
 * `i = 2`, error `-FDT_ERR_NOTFOUND`
 * `i = 3`, error `-FDT_ERR_BADVALUE`
 * `i = 4`, error `-FDT_ERR_BADPATH`

Looking at it from a user's perspective: The user asks for a property at the path "sa3". Replying with

    FDT_ERR_BADPATH: Function was passed a badly formatted path

seems right to me and also matches the behavior of fdtget before your changes(!). However replying with

    FDT_ERR_BADVALUE: Device tree has a property with an unexpected value.

(while being correct) is a diagnosis about the property `/aliases/sa3` and not about the path "sa3" that the user provided and so only something like an implementation detail.

I'm not authoritative here, so I'll stop arguing here and let @dgibson think and decide about that.

libfdt/libfdt.h
ukleinek: Ah indeed. It's documented and I missed that. While working on dtc for my change I stumbled about this "feature" that is inherited to e.g. `fdt_overlay_target_offset()` which is surprising at least. (If you're subscribed to `devicetree-compiler@vger.kernel.org`, see my mail "Bug or feature? /somenode overwrites /somenode@17 in fdt-overlay" with Message-Id `20230425145504.tswuitufzikuhmnz@pengutronix.de`, I didn't find it in a web archive.)

Villemoes: >  (If you're subscribed to `devicetree-compiler@vger.kernel.org`, see my mail "Bug or feature? /somenode overwrites /somenode@17 in fdt-overlay" with Message-Id `20230425145504.tswuitufzikuhmnz@pengutronix.de`, I didn't find it in a web archive.)

I'm not subscribed, but it seems that that ML should be archived on lore.kernel.org, and if it isn't already (I can't find it), someone should ask helpdesk to do it (see https://korg.docs.kernel.org/lore.html).

libfdt/fdt_ro.c
dgibson: I don't think we should extend the semantics of `/aliases`.  That's been around since the ancient history of OF, and I think extending it is of very little use here.  Noting that you can already do:

```
/ {
        aliases {
                sa = &label;
        }
}
```
Where dtc will expand that to the full path.

I'm not 100% sure, but I believe it's also valid for aliases to reference other aliases.  I couldn't quickly find absolute confirmation in IEEE1275, but I think this was valid an used in practice in the old OF days.  For example:
```
/ {
        aliases {
                sa1 = "/foo/bar";
                sa2 = "sa1/baz";
        };
}
````

File: libfdt/fdt_ro.c (hunk 6 above)
Comment (dgibson) [reply libfdt/fdt_ro.c]: `path` would be better, matching `fdt_path_offset()`.

File: libfdt/fdt_ro.c (hunk 5 above)
Comment (Villemoes) [reply libfdt/fdt_ro.c:274]: > > I'm not 100% sure, but I believe it's also valid for aliases to reference other aliases. I couldn't quickly find absolute confirmation in IEEE1275, but I think this was valid and used in practice in the old OF days. For example:
> > ```
> > / {
> >         aliases {
> >                 sa1 = "/foo/bar";
> >                 sa2 = "sa1/baz";
> >         };
> > }
> > ```
> 
> Well, if that's really the case, then of course my sanity check patch should be removed. 

Ah, but https://www.devicetree.org/specifications/ is quite clear:
```
Each property of the ``/aliases`` node defines an alias. The property name
specifies the alias name. The property value specifies the full path to
a node in the devicetree.
```
and
```
An alias value is a device path and is encoded as a string. The value
represents the full path to a node, but the path does not need to refer
to a leaf node.
```
That text seems to be essentially unchanged since it first appeared in that git repo.

dgibson: > > > I'm not 100% sure, but I believe it's also valid for aliases to reference other aliases. I couldn't quickly find absolute confirmation in IEEE1275, but I think this was valid and used in practice in the old OF days. For example:
> > > ```
> > > / {
> > >         aliases {
> > >                 sa1 = "/foo/bar";
> > >                 sa2 = "sa1/baz";
> > >         };
> > > }
> > > ```
> > 
> > 
> > Well, if that's really the case, then of course my sanity check patch should be removed.
> 
> Ah, but https://www.devicetree.org/specifications/ is quite clear:

Eh.. maybe..
 
> ```
> Each property of the ``/aliases`` node defines an alias. The property name
> specifies the alias name. The property value specifies the full path to
> a node in the devicetree.
> ```
> 
> and
> 
> ```
> An alias value is a device path and is encoded as a string. The value
> represents the full path to a node, but the path does not need to refer
> to a leaf node.
> ```

So, both of these look to be paraphrased from IEEE1275.  The ambiguity, such as it is, is in the meaning of "full path".  1275 uses the term `device-path` in the equivalent place, and I think that term may allow the first component to be an alias, rather than `/`, though I'm having trouble pinning down something which says that one way or another.  I do see `device-path` used in other contexts of 1275 where I'm pretty certain aliases are allowed (e.g. the value of the `bootpath` property of `/chosen`).

> That text seems to be essentially unchanged since it first appeared in that git repo.

As you'd expect for something more or less inherited from 1275.

dgibson:
On Fri, Apr 28, 2023 at 01:15:23AM -0700, Villemoes wrote:
> @Villemoes commented on this pull request.
> 
> 
> 
> > +		if (!can_assume(VALID_INPUT) && *p != '/')
> +			return -FDT_ERR_BADVALUE;
> +
> 
> > > I'm not 100% sure, but I believe it's also valid for aliases to reference other aliases. I couldn't quickly find absolute confirmation in IEEE1275, but I think this was valid and used in practice in the old OF days. For example:
> > > ```
> > > / {
> > >         aliases {
> > >                 sa1 = "/foo/bar";
> > >                 sa2 = "sa1/baz";
> > >         };
> > > }
> > > ```
> > 
> > Well, if that's really the case, then of course my sanity check patch should be removed. 
> 
> Ah, but https://www.devicetree.org/specifications/ is quite clear:

Eh.. not necessarily.

> ```
> Each property of the ``/aliases`` node defines an alias. The property name
> specifies the alias name. The property value specifies the full path to
> a node in the devicetree.
> ```
> and
> ```
> An alias value is a device path and is encoded as a string. The value
> represents the full path to a node, but the path does not need to refer
> to a leaf node.
> ```

So, both of these look to be paraphrased from IEEE1275.  The
ambiguity, such as it is, is in the meaning of "full path".  1275 uses
the term "device-path" in the equivalent place, and I think that term
may allow the first component to be an alias, rather than '/', though
I'm having trouble pinning down something which says that one way or
another.  I do see "device-path" used in other contexts of 1275 where
I'm pretty certain aliases are allowed (e.g. the value of the
'bootpath' property of /chosen).

> That text seems to be essentially unchanged since it first appeared in that git repo.

As you'd expect for something more or less inherited from 1275.

-- 
				| _way_ _around_!

libfdt/fdt_ro.c
Villemoes: > The ambiguity, such as it is, is in the meaning of "full path". 

Alright, I've pushed a version without that sanity check (and with s/nodename/path/).

dgibson:
I've merged the first 4 patches.  I need to think a bit more about the remaining two.

Villemoes:
I have rebased the last two patches on top of current main. Can they be considered for inclusion?

PR #92, opened 2023-04-14
Fix segmentation fault in check_graph_node and crash in check_graph_child_address

jbeisswenger-cetitec:
`check_graph_node()` dereferences `node->parent` without checking whether it is `NULL` first (i.e. the root node).
As a result a segmentation fault occurs for dts files which contain an 'endpoint' node as a direct child of the
root node. This type of error can easily happen when a 'remote-endpoint' property is
accidentally placed outside the corresponding endpoint and port nodes.

Minimal example with 'endpoint' node:
```
/dts-v1/;
/ {	endpoint {};  };
```

Minimal example with remote-endpoint property:
```
/dts-v1/;
/ {
	foo {
                remote-endpoint = <0xdeadbeef>;
	};
};
```
While fixing above issue I ran into an assertion in `check_graph_child_address` which can be triggered with the following example:

```
/dts-v1/;
 / {
	bar: bar {
		port {
			bar_con: endpoint {
				remote-endpoint = <&foo_con>;
			};
		};
	};
	foo {
		port {
			#address-cells = <1>;
			#size-cells = <1>; // should always be 0
			foo_con: endpoint@1 {
				reg = <1 2>; // causes assertion failure instead of diagnostic
				remote-endpoint = <&bar_con>;
			};
		};
	};
};
```

The condition that triggers this issue is actually caught by `check_graph_port()`, however `check_graph_child_address()` runs first hiding the warning.

### DIFF
--- a/checks.c
+++ b/checks.c
@@ -1785,31 +1785,6 @@ static void check_graph_nodes(struct check *c, struct dt_info *dti,
 }
 WARNING(graph_nodes, check_graph_nodes, NULL);
 
-static void check_graph_child_address(struct check *c, struct dt_info *dti,
-				      struct node *node)
-{
-	int cnt = 0;
-	struct node *child;
-
-	if (node->bus != &graph_ports_bus && node->bus != &graph_port_bus)
-		return;
-
-	for_each_child(node, child) {
-		struct property *prop = get_property(child, "reg");
-
-		/* No error if we have any non-zero unit address */
-		if (prop && propval_cell(prop) != 0)
-			return;
-
-		cnt++;
-	}
-
-	if (cnt == 1 && node->addr_cells != -1)
-		FAIL(c, dti, node, "graph node has single child node '%s', #address-cells/#size-cells are not necessary",
-		     node->children->name);
-}
-WARNING(graph_child_address, check_graph_child_address, NULL, &graph_nodes);
-
 static void check_graph_reg(struct check *c, struct dt_info *dti,
 			    struct node *node)
 {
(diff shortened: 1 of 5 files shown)


checks.c
```diff
@@ -1798,8 +1798,10 @@ static void check_graph_child_address(struct check *c, struct dt_info *dti,
 		struct property *prop = get_property(child, "reg");
 
 		/* No error if we have any non-zero unit address */
-		if (prop && propval_cell(prop) != 0)
+		/* We have a check for val.len == sizeof(cell_t) elsewhere */
```
dgibson, line 1801: If that's the case, can't you move that above this point and then you could safely use `propval_cell`

checks.c
```diff
@@ -1798,8 +1798,10 @@ static void check_graph_child_address(struct check *c, struct dt_info *dti,
...
+                if (prop && propval_cell_n(prop, 0)  != 0 ) {
```
dgibson, line 1802: Do you have a danger of a different assertion failure if the length of prop is < 4?

There's also an extraneous space here, before the `!=`.

checks.c
```diff
@@ -1798,8 +1798,10 @@ static void check_graph_child_address(struct check *c, struct dt_info *dti,
...
 			return;
+                }
```
dgibson, line 1804: The style I'm using for dtc does not include braces around single line if bodies.

dgibson (commented):
I've merged the first two patches, but I have some quieries about the third.

Sorry its taken me so long to look at this.
b

checks.c
jbeisswenger-cetitec: Since the case of prop length < 4 cannot simply be handled by using `propval_cell_n` (with the assertion fixed) instead of `propval_cell` I am reverting this change. Instead I'm adding `graph_port` and `graph_endpoint` as prerequisites for `graph_child_address`.

Another alternative would be to call `check_graph_reg()` in `check_graph_child_address()` directly but then any warnings in `check_graph_reg()` would be issued twice.

jbeisswenger-cetitec (commented):
Thank you for your review!

I'm reverting the 3rd commit and instead I'm adding the `graph_endpoint` and `graph_port` checks as prerequisites for the `graph_child_address` check which triggered the assertion.

jbeisswenger-cetitec:
Rebased onto main.

checks.c
dgibson: > Edit: `check_graph_port()` and `check_graph_endpoint()` are now prerequisites of `check_graph_child_address()` and `propval_cell` is used again.

That sounds like the right solution: this is exactly the sort of thing that the prerequisite system is for.

dgibson:
Thanks for the update, this looks good now.  Merged.

PR #94, opened 2023-04-28
Convert CI definitions to use GitHub Actions

ukleinek:
This drops the old CI definitonis for Travis and Cirrus and replaces it by GitHub Actions instead.

Building dtc with libyaml older than 0.2.3 results in various compiler warnings that make the CI pipeline fail. So a commit is added to depend on a new enough version of libyaml.


ukleinek:
Side node: This pull request isn't automatically tested yet because the CI definitions are missing in dgibson/dtc:main. You can see a successfull run at https://github.com/ukleinek/dtc/actions/runs/4829388801

Makefile
```diff
@@ -55,7 +55,11 @@ else
 	CFLAGS += $(shell $(PKG_CONFIG) --cflags valgrind)
```
dgibson, line 55: Patch looks fine, but you're missing a signed-off-by line.

.github/workflows/build.yml
```diff
@@ -0,0 +1,33 @@
...
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
```
dgibson, line 16: Is it possible to also run this on other distros?  Testing on a variety would be quite valuable.

dgibson (changes requested):
In addition to the inline comments, some patches are missing signed-off-by lines, and the "drop obsolete.." patch needs some content in the commit message, describing the background of those CI systems dissapearing.

Makefile
ukleinek: Added a Sob-lines for all commits now. I suggest installing https://github.com/apps/dco for your repo, this adds an automatic check for missing Sob-lines.

.github/workflows/build.yml
ukleinek: I think Ubuntu is the only (freely) available distro. You can pick older versions though.
See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

Also adding custom runners should be possible, but I don't know how that works.

Biswa96: It is possible to test with multiple distribution using docker. I can try to help with that after this is merged.

dgibson: Ok.  It's a pity it's not easier to test other things, but something is better than nothing.  @Biswa96 that would be most appreciated i fyou could extend the coverage in future.

dgibson (commented):
Looks good to me, at least as far as I can tell not really knowing anything about github actions.

Biswa96:
I have tried to use containers with GitHub actions. Only Ubuntu succeed. All other distributions archl, alpine, fedora failed with this error.

```
LookupError: setuptools-scm was unable to detect version for /__w/dtc/dtc.
```

Any idea how to troubleshoot that error?

Biswa96:
I have tried to use containers with GitHub actions. It failed with this error.

```
LookupError: setuptools-scm was unable to detect version for /__w/dtc/dtc.
```

Any idea how to troubleshoot that error?

Biswa96:
One issue. The `.git` directory is not present when using container in GitHub Actions. That shows the following error.

```
LookupError: setuptools-scm was unable to detect version for /__w/dtc/dtc.
```

PR #95, opened 2023-05-02
Fix version in libfdt/meson.build

anatol:
Currently meson generates file /usr/lib/libfdt.so.1.6.0 that does not match Makefile build version.

### DIFF
--- a/libfdt/meson.build
+++ b/libfdt/meson.build
@@ -18,7 +18,7 @@ sources = files(
 
 libfdt = library(
   'fdt', sources,
-  version: '1.6.0',
+  version: '1.7.0',
   link_args: ['-Wl,--no-undefined', version_script],
   link_depends: 'version.lds',
   install: true,


dgibson:
Ugh, yes.  I already knew of the duplicated version between `Makefile` and `meson.build`, but I hadn't spotted there was a third version in `libfdt/meson.build` in the specifics for the shared library.  We really need to remove this redundancy.

I suspect there's a nice way to do that in meson, but I don't know what it is.

dgibson (changes requested):
So medium to long term, I'd really like a structural solution to this - somehow obtaining both the Makefile and meson versions form the same source so they can't get out of sync.

In the meantime this looks like a correct quick fix, however, I'll need a Signed-off-by line in order to apply it (see `CONTRIBUTING.md`).

eli-schwartz:
Since the library soname is always based on the project release version, you can just use `version: meson.project_version()` to ensure that `meson.build` never gets out of sync with `libfdt/meson.build`.

Syncing the Makefile and meson.build is a different question. As far as I can tell, in the Makefile build you never actually use the individual integer components, so you could just store the version in a file as `DTC_VERSION = $(shell cat VERSION.txt)`. For meson, that can be read in with
```meson
project(
    ...,
    version: files('VERSION.txt'),
    meson_version: '>=0.57.0'
)
```

Note your current minimum required meson is not spelled out but is 0.56.0 due to use of features introduced in that version. Bumping from 0.56.0 to 0.57.0 is a *bit* notable inasmuch as Debian stable ships 0.56.2 unless users enable the backports repository, in which case they receive 1.0.0.

libfdt/meson.build
```diff
@@ -18,7 +18,7 @@ sources = files(
 
 libfdt = library(
   'fdt', sources,
-  version: '1.6.0',
+  version: '1.7.0',
```
ernestask, line 21:
```suggestion
  version: meson.project_version(),
```

ernestask:
This is causing me issues somewhat. @anatol, do you plan to continue with this PR?

blmaier:
@anatol @dgibson this was fixed awhile ago, this pull request can be closed

dgibson:
@blmaier thanks for the update.

PR #96, opened 2023-05-14
CI: Add build matrix with multiple Linux distributions

Biswa96:


dgibson:
> `make check` fails in Alpine Linux. I do not have the expertise to debug the test failures. If you allow I can ask some experts.

I'm afraid I don't know how to debug this either, at least not quickly.  Feel free to ask whoever you need.

In the meantime, could you modify this series so it adds the non-alpine distros, then we can get at least some better coverage until we debug the alpine problem.

dgibson:
> I figured out the test failure in Alpine Linux. I missed the coreutils packages as provided here https://git.alpinelinux.org/aports/tree/main/dtc/APKBUILD.

Tremendous!

Merged.

PR #98, opened 2023-05-16
overlay: ensure that existing phandles are not overwritten

ukleinek:
A phandle in an overlay is not supposed to overwrite a phandle that already exists in the base dtb as this breaks references to the respective node in the base.

Traditionally all phandle values in the overlay were increased by a fixed offset such that the base and overlay handle values don't overlap. With this change the overlay's phandles are determined dynamically. Already existing phandles are kept and an array "phandlemap" is used to track how the overlay's phandles change.

overlay_adjust_local_phandles() which creates the mapping between old and new phandles in the overlay needs to check the nodes in the base. To be able to find the base's node, overlay_fixup_phandles() must be run first, so the order of functions fdt_overlay_apply() had to change.

A test is added that checks that newly added phandles and existing phandles work as expected.

### DIFF
--- a/libfdt/fdt_overlay.c
+++ b/libfdt/fdt_overlay.c
@@ -85,42 +85,38 @@ int fdt_overlay_target_offset(const void *fdt, const void *fdto,
 }
 
 /**
- * overlay_phandle_add_offset - Increases a phandle by an offset
+ * overlay_phandle_set - Set a phandle to a specific value
  * @fdt: Base device tree blob
  * @node: Device tree overlay blob
  * @name: Name of the property to modify (phandle or linux,phandle)
- * @delta: offset to apply
+ * @phandleval: value to set the phandle to
+ * @phandlemap: List to map previous phandle values to new ones.
  *
- * overlay_phandle_add_offset() increments a node phandle by a given
- * offset.
+ * overlay_phandle_set() set a node's phandle to a given
+ * value and updates phandlemap accordingly.
  *
  * returns:
  *      0 on success.
  *      Negative error code on error
  */
-static int overlay_phandle_add_offset(void *fdt, int node,
-				      const char *name, uint32_t delta)
+static int overlay_phandle_set(void *fdt, int node,
+			       const char *name, uint32_t phandleval, uint32_t *phandlemap)
 {
-	const fdt32_t *val;
-	uint32_t adj_val;
 	int len;
+	const fdt32_t *phandle;
+	uint32_t prevval;
 
-	val = fdt_getprop(fdt, node, name, &len);
-	if (!val)
+	phandle = fdt_getprop(fdt, node, name, &len);
+	if (!phandle)
 		return len;
 
-	if (len != sizeof(*val))
+	if (len != sizeof(*phandle))
 		return -FDT_ERR_BADPHANDLE;
 
-	adj_val = fdt32_to_cpu(*val);
-	if ((adj_val + delta) < adj_val)
-		return -FDT_ERR_NOPHANDLES;
+	prevval = fdt32_ld_(phandle);
+	phandlemap[prevval - 1] = phandleval;
 
-	adj_val += delta;
-	if (adj_val == (uint32_t)-1)
-		return -FDT_ERR_NOPHANDLES;
-
-	return fdt_setprop_inplace_u32(fdt, node, name, adj_val);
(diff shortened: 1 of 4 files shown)


ukleinek:
Bah, the compiler on arch and fedora is stupid. I don't have a nicer idea than to initialize `phandle`.

pelwell:
Yes, a scheme that preserves existing phandles rather than inventing new ones is ideal, allowing for both functionalities (intentionally adding a new label to a base dtb, and applying an overlay which just happens to have a label that will clash) but it does mean that the application step becomes more involved.

> Bah, the compiler on arch and fedora is stupid. I don't have a nicer idea than to initialize phandle.

I don't understand what has provoked this comment - the automatic checks are happy.

ukleinek:
> Yes, a scheme that preserves existing phandles rather than inventing new ones is ideal, allowing for both functionalities (intentionally adding a new label to a base dtb, and applying an overlay which just happens to have a label that will clash) but it does mean that the application step becomes more involved.

If code has a bug because it makes wrong assumptions, it's quite usual that there is an extra effort to make it right.

> > Bah, the compiler on arch and fedora is stupid. I don't have a nicer idea than to initialize phandle.
> 
> I don't understand what has provoked this comment - the automatic checks are happy.

They are, but only after https://github.com/dgibson/dtc/compare/e8db66a5580ba91f1552406ec1fc73ad25108006..42b2aab5b6f182541ee1524395d642a3c829c6d2 .

dgibson:
Sorry.  One of the design constraints for libfdt is that it needs to work in constrained environments that don't have an allocator, which this patch breaks.  That's exactly why we're using the kind of clunky means of adding a fixed offset.

libfdt/fdt_overlay.c
```diff
@@ -816,23 +884,49 @@ int fdt_overlay_apply(void *fdt, void *fdto)
...
+	uint32_t *phandlemap = NULL;
 
 	FDT_RO_PROBE(fdt);
 	FDT_RO_PROBE(fdto);
 
+	ret = fdt_find_max_phandle(fdto, &delta);
+	if (ret)
+		goto err;
+
+	phandlemap = calloc(delta, sizeof(*phandlemap));
```
dgibson, line 896: Note that even if allocations were ok in general, this could easily be a *huge* allocation.  If whatever created the overlay happened to use a single large phandle value (which is valid, though unusual) you could allocate up to 16GiB here.

ukleinek:
> Sorry. One of the design constraints for libfdt is that it needs to work in constrained environments that don't have an allocator, which this patch breaks. That's exactly why we're using the kind of clunky means of adding a fixed offset.

The bug however is real. If allocating memory isn't an option, then the only way to fix it would be to merge the functionality of `overlay_adjust_local_phandles()` and `overlay_update_local_references()` into a single function and iterate over `__local_fixups__` once for each phandle value. Would you be open to such a change even though it increases runtime complexity? Or do you have a better idea?

dgibson:
> > Sorry. One of the design constraints for libfdt is that it needs to work in constrained environments that don't have an allocator, which this patch breaks. That's exactly why we're using the kind of clunky means of adding a fixed offset.
> 
> The bug however is real. If allocating memory isn't an option, then the only way to fix it would be to merge the functionality of `overlay_adjust_local_phandles()` and `overlay_update_local_references()` into a single function and iterate over `__local_fixups__` once for each phandle value. Would you be open to such a change even though it increases runtime complexity? Or do you have a better idea?

Yes, I'm open to such a change, even though it's certainly more complex than I'd ideally like.  I had some more thoughts about the bug in #97 (here probably would have made more sense in hindsight).

ukleinek:
> > Sorry. One of the design constraints for libfdt is that it needs to work in constrained environments that don't have an allocator, which this patch breaks. That's exactly why we're using the kind of clunky means of adding a fixed offset.
> 
> The bug however is real. If allocating memory isn't an option, then the only way to fix it would be to merge the functionality of `overlay_adjust_local_phandles()` and `overlay_update_local_references()` into a single function and iterate over `__local_fixups__` once for each phandle value. Would you be open to such a change even though it increases runtime complexity? Or do you have a better idea?

I thought a bit about this, and the idea is buggy. We first have to implement the fixed shift in the overlay and then in an additional pass check for conflicts and fixup. Otherwise it might happen that I update all refs with value 3 in the overlay to 5 to match the base dtb and then modify these again when handling the overlay refs with value 5.

PR #101, opened 2023-05-31
libfdt: use modern include guards

axel-h:


a3f:
`#pragma once` is a non-standard extension. I think you need a better rationale why this change should be accepted.

dgibson:
Right, I'm afraid I've never even heard of these so called "modern guards".  I'm not willing to use a `pragma` for this.

PR #102, opened 2023-05-31
libfdt: meson: Fix linking on macOS linker

otavio:
-undefined error is the equivalent of --no-undefined for the macOS linker, but -undefined would also be understood as a valid argument for GNU ld so we use the supported linker variant.


dgibson:
Merged, thanks.

PR #103, opened 2023-06-01
meson: allow disabling tests

otavio:


dgibson (changes requested):
This needs a much more substantial commit message.  What does this allow that's different from just not running `meson test`?

alyssais (changes requested):
This would make the tests be enabled by default when cross compiling, which would cause Meson to fail to configure.

eli-schwartz:
Failing to configure sounds odd -- unless you don't have an exe_wrapper configured (wine, qemu-user, etc. for running cross-built executables against cross-built libraries).

alyssais:
> unless you don't have an exe_wrapper configured

That's what I mean, indeed.  (We don't have an exe wrapper configured by default in Nixpkgs, because qemu is a big build dependency, and we don't run tests for cross builds anyway.)

elmarco (approved):

dgibson:
Merged, thanks.

PR #105, opened 2023-07-27
Fix use of <ctype.h> functions

### LINKED ISSUE #71: Char subscript error in Cygwin

sebhub:
The value passed to the <ctype.h> functions shall be the value of an unsigned char or EOF.  It is implementation-defined if the char type is signed or unsigned.  Cast to unsigned char to avoid undefined behaviour on systems where char is signed.

This cast is already present in other parts of the code base.

Close #71.

### DIFF
--- a/checks.c
+++ b/checks.c
@@ -1182,7 +1182,7 @@ static void check_unit_address_format(struct check *c, struct dt_info *dti,
 		/* skip over 0x for next test */
 		unitname += 2;
 	}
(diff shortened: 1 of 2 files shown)


elmarco (approved):

dgibson:
It seems pretty nuts to me that you can't safely use functions explicitly designed to test characters on values of type `char`.  But I guess that's the weird history of the C library for you.

Merged.

PR #108, opened 2023-07-28
libfdt: Add symbols to their corresponding versions

ernestask:
787b599c8108f514645d9ee6256ee53dccc9db84 added the linker script with symbol versions, but, over time, new versions were not added, making versioning somewhat pointless. This commit aims to put symbols under versions they were introduced in. As a result, some previously-local symbols also become global.

### DIFF
--- a/libfdt/version.lds
+++ b/libfdt/version.lds
@@ -1,33 +1,13 @@
 /* SPDX-License-Identifier: (GPL-2.0-or-later OR BSD-2-Clause) */
-LIBFDT_1.2 {
+LIBFDT_1.0 {
 	global:
-		fdt_next_node;
-		fdt_check_header;
 		fdt_move;
 		fdt_string;
-		fdt_num_mem_rsv;
-		fdt_get_mem_rsv;
 		fdt_subnode_offset_namelen;
 		fdt_subnode_offset;
-		fdt_path_offset_namelen;
 		fdt_path_offset;
-		fdt_get_name;
-		fdt_get_property_namelen;
 		fdt_get_property;
-		fdt_getprop_namelen;
 		fdt_getprop;
-		fdt_get_phandle;
-		fdt_get_alias_namelen;
-		fdt_get_alias;
-		fdt_get_path;
-                fdt_header_size;
-		fdt_supernode_atdepth_offset;
-		fdt_node_depth;
-		fdt_parent_offset;
-		fdt_node_offset_by_prop_value;
-		fdt_node_offset_by_phandle;
-		fdt_node_check_compatible;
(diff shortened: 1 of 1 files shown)


libfdt/version.lds
```diff
@@ -20,7 +20,7 @@ LIBFDT_1.2 {
 		fdt_get_alias_namelen;
 		fdt_get_alias;
 		fdt_get_path;
-                fdt_header_size;
+		fdt_header_size;
```
ernestask, line 23: This was using spaces instead of tabs, so I fixed it in the same commit.

ernestask:
Gah, I put them under the wrong version.

ernestask:
I’ve changed the topic of this PR and shuffled symbols around to match the versions they were introduced in, as now all of them will appear under `LIBFDT_1.2`. Wasn’t sure what to do with unreleased ones, so I put them under a fictitious 1.8 version.

dgibson:
> [787b599](https://github.com/dgibson/dtc/commit/787b599c8108f514645d9ee6256ee53dccc9db84) added the linker script with symbol versions, but, over time, new versions were not added, making versioning somewhat pointless.

Uh... yes and no.  So, my understanding is that *adding* new symbols to an existing version is more or less safe.  It's perhaps not the ideal way of doing things, but it will work: old clients expecting the old version obviously won't be using the new symbols, but newer programs expecting the newer versions will find them.  We only break the symbol versioning if we *change* the signature for a symbol without adding a new symbol version.

> This commit aims to put symbols under versions they were introduced in. As a result, some previously-local symbols also become global.

While it probably would have been neater if we'd done it this way all along, I think changing this now could break existing binaries using the shared library: a binary could have been built with a recentish libfdt (but before this change) so it is expecting to see a new symbol (say `fdt_overlay_target_offset()`) within the `LIBFDT_1.2` version.  That would then break when it tries to run against a shared library with this change, where that symbol has been moved to a newer version.

So while it would be reasonable to always add new symbols to new versions *in future*, I don't think we can move already versioned symbols around.

libfdt/version.lds
```diff
@@ -40,44 +20,99 @@ LIBFDT_1.2 {
...
+		fdt_resize;
 		fdt_address_cells;
 		fdt_size_cells;
-		fdt_stringlist_contains;
+		fdt_path_offset_namelen;
 		fdt_stringlist_count;
 		fdt_stringlist_search;
 		fdt_stringlist_get;
-		fdt_resize;
+		fdt_get_max_phandle;
```
blmaier, line 91: This symbol was changed into a header file inline in 1.6.0

PR #109, opened 2023-07-28
pylibfdt: Use python.extension_module()

ernestask:
This commit avoid using deprecated setuptools utilities and builds the module directly using Meson.

### DIFF
--- a/pylibfdt/meson.build
+++ b/pylibfdt/meson.build
@@ -1,12 +1,19 @@
-setup_py = find_program('../setup.py')
-setup_py = [setup_py.path(), '--quiet', '--top-builddir', meson.project_build_root()]
-
-custom_target(
-  'pylibfdt',
-  input: 'libfdt.i',
-  output: '_libfdt.so',
-  command: [setup_py, 'build_ext', '--build-lib=' + meson.current_build_dir()],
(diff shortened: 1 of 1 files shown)


ernestask:
I did some quick sanity checking by replacing the module and running `make check`, but it turns out that https://github.com/dgibson/dtc/pull/108 was causing tests to not run.

dgibson:
With this patch meson builds fail for me with lots of errors from swig.

We also still get warnings when building with make because of deprecated setuptools.  I'd prefer to avoid diverging how the module is built between make and meson if we can.

ernestask:
The thing about build systems is that they are not supposed to mimic one another. If you want to use Meson like Make, then it’s best to drop support for it.

ernestask:
But feel free to close this if you disagree, I can maintain a fork with these changes.

dgibson:
> The thing about build systems is that they are not supposed to mimic one another. If you want to use Meson like Make, then it’s best to drop support for it.

It's not a question of wanting to use meson like make, it's a question of (confidently) producing similar results.  What I'm worried about here is the possibility of subtle incompatibilities in the python package as produced by meson versus as produced by setuptools.

PR #110, opened 2023-07-28
workflows: build: Add Meson build job

ernestask:
This commit adds a parallel job to run Meson test builds.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ name: Build test
       - main
 
 jobs:
-  build:
+  build-make:
     runs-on: ubuntu-latest
 
     strategy:
@@ -35,3 +35,25 @@ jobs:
       - name: Run check
         run: |
           SETUPTOOLS_SCM_PRETEND_VERSION="0" make check
+  build-meson:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+          os: [ "alpine", "archlinux", "fedora", "ubuntu" ]
+
+    container:
+      image: ${{ matrix.os }}
+
+    steps:
+      - name: Check out source
+        uses: actions/checkout@v3
+
+      - name: Install dependencies
+        run: "${GITHUB_WORKSPACE}/scripts/install-deps.sh"
+
+      - name: Build
+        run: |
+          meson setup . _build
+          meson compile -C _build
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,11 +43,27 @@ jobs:
       fail-fast: false
       matrix:
           os: [ "alpine", "archlinux", "fedora", "ubuntu" ]
+          include:
+            - install_git_cmd: apk add git
+              os: alpine
+            - install_git_cmd: pacman -Sy --noconfirm git
+              os: archlinux
+            - install_git_cmd: dnf --assumeyes install git
+              os: fedora
+            - install_git_cmd: |
+                export DEBIAN_FRONTEND=noninteractive
+                apt-get update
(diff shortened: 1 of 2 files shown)


ernestask:
Seems like a misplaced `-Werror` is causing the Python module to not build (SWIG spits out some not-nice code).

dgibson:
@ernestask in the github action itself I'm seeing setuptools errors before we even get to swig.

ernestask:
Oh, right, that is caused by not having a recent-enough git. If you check the logs for the checkout step, you’ll see a warning.

ernestask:
But this is a draft anyway, I’ll continue tinkering today.

ernestask:
A couple of hacks later, it seems to be building.

- setuptools_scm requires a git repo, but not having git installed when using `actions/checkout` results in a source archive being fetched and exploded. I fixed that by adding a step before checkout to install git.

- Later, Ubuntu builds were failing as described in https://github.com/actions/checkout/issues/766, so another step was required to add $GITHUB_WORKSPACE to safe.directory in .gitconfig.
 
Make builds are not affected, because they use a different kludge (`SETUPTOOLS_SCM_PRETEND_VERSION="0"`).

dgibson:
> Oh, right, that is caused by not having a recent-enough git. If you check the logs for the checkout step, you’ll see a warning.

Uh.. not recent enough which git?  I'm literally working from the dtc main branch...

ernestask:
> Uh.. not recent enough which git? I'm literally working from the dtc main branch...

I mean in the Actions runner. Currently you use a base container image for multiple distros, none of which have git installed.

dgibson:
> > Uh.. not recent enough which git? I'm literally working from the dtc main branch...
> 
> I mean in the Actions runner. Currently you use a base container image for multiple distros, none of which have git installed. My phrasing could have been better.

Ah!  Now I see.

SoldierC4:
> A couple of hacks later, it seems to be building.
> 
> * setuptools_scm requires a git repo, but not having git installed when using `actions/checkout` results in a source archive being fetched and exploded. I fixed that by adding a step before checkout to install git.
> * Later, Ubuntu builds were failing as described in [fatal: unsafe repository (REPO is owned by someone else) in other workflow steps after running checkout actions/checkout#766](https://github.com/actions/checkout/issues/766), so another step was required to add $GITHUB_WORKSPACE to safe.directory in .gitconfig.
> 
> Make builds are not affected, because they use a different kludge (`SETUPTOOLS_SCM_PRETEND_VERSION="0"`).

I opened a pull request to fix the setuptools_scm issue so a git repo is no longer required https://github.com/dgibson/dtc/pull/114

.github/workflows/build.yml
```diff
@@ -35,3 +36,45 @@ jobs:
...
+      - name: Install dependencies
+        run: "${GITHUB_WORKSPACE}/scripts/install-deps.sh"
+
+      # https://github.com/actions/checkout/issues/766
+      - name: Add workspace to safe.directory
+        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+
+      - name: Build
+        run: |
+          meson setup . _build
```
SoldierC4, line 79: Once Meson deprecations are fixed in https://github.com/dgibson/dtc/pull/113, we can use `meson setup --fatal-meson-warnings` to enforce version warnings.

dgibson:
@ernestask as @blmaier notes we have some fixes on the way which should make this worth rebasing.

Note that before I can apply, I'll also need at least Signed-off-by lines on each of the patches in this series.  Most of the patches will also need a more substantial commit message giving the rationale for each change (just a subject line is almost never sufficient).

PR #112, opened 2023-07-29
libfdt: fix library version to match project version

SoldierC4:
Build the libfdt with the correct version number by pulling the version from the top-level project.

Change as suggested from https://github.com/dgibson/dtc/pull/95#issuecomment-1546933095

### DIFF
--- a/libfdt/meson.build
+++ b/libfdt/meson.build
@@ -28,7 +28,7 @@ endif
 link_args += version_script
 libfdt = library(
   'fdt', sources,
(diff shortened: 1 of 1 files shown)


dgibson:
Applied.  Note however, that this is only half of the change suggested at https://github.com/dgibson/dtc/pull/95#issuecomment-1546933095 .  This makes the library version match the project version, but doesn't load the project version from VERSION.txt which would ensure that the Makefile and meson builds keep in sync with version.

PR #114, opened 2023-08-02
move release version into VERSION.txt

blmaier:
> move release version into VERSION.txt
> 
> To synchronize the release version of the Makefile and Meson build
> systems, pull the version info from a shared file.
> 
> When building the Meson shared library, we strip off the optional '-rc1'
> suffix as Meson requires library versions use only X.Y.Z versioning.

This increases the minimum Meson version from 0.56.0 to 0.57.0. Notably 0.56.0 is the version Debian buster-backports and Debian bullseye still use.
https://repology.org/project/meson/versions

Implements suggestion from https://github.com/dgibson/dtc/pull/95#issuecomment-1546933095

> pylibfdt: use fallback version in tarballs
> 
> When building pylibfdt from the released tarballs[1] setup.py will fail
> with the following.
> 
> > LookupError: setuptools-scm was unable to detect version for dtc.
> > Make sure you're either building from a fully intact git repository or
> > PyPI tarballs. Most other sources (such as GitHub's tarballs, a git
> > checkout without the .git folder) don't contain the necessary metadata
> > and will not work.
> 
> seutptools_scm supports a 'fallback_version' that will use the provided
> string as the version when the git repo is not available.
> 
> [1] https://www.kernel.org/pub/software/utils/dtc/dtc-1.7.0.tar.xz

> workflows: build: remove setuptools_scm hack
> 
> The SETUPTOOLS_SCM_PRETEND_VERSION was needed as ./setup.py would fail
> without it. As setuptools_scm will fail if there is not git repo, and
> the github workflow container does not include the source code git repo.
> 
> A previous commit added "fallback_version" to setuptools_scm which
> instructs it to use the version from VERSION.txt when the git repo is
> missing. So this hack is no longer needed.


libfdt/meson.build
```diff
@@ -28,7 +28,7 @@ endif
 link_args += version_script
 libfdt = library(
   'fdt', sources,
-  version: meson.project_version(),
+  version: meson.project_version().split('-')[0],
   link_args: link_args,
```
dgibson, line 32: Does meson use the version here to control the soname as well as the name of the binary?  If so, that's a bit concerning, because it means the meson built and make built sonames are already different, which could lead to incompatibilities in the output.

dgibson (commented):
These all look good in outline.  I do have one query noted here.

Note that all patches will need a S-o-b line before I can merge.

libfdt/meson.build
blmaier: Meson uses the major version number for the soname, so they end up the same. Below is examining the soname with the an "-rc1" release candidate added to the VERSION.txt.

Makefile
```
$ objdump -p libfdt/libfdt.so.1 | grep 'SONAME\|LIBFDT'
  SONAME               libfdt.so.1
2 0x00 0x0a9d2f72 LIBFDT_1.2
```

Meson
```
$ objdump -p build/libfdt/libfdt.so.1 | grep 'SONAME\|LIBFDT'
  SONAME               libfdt.so.1
2 0x00 0x0a9d2f72 LIBFDT_1.2
```

**However** the final shared library names are different. But in this case at least I believe it does not matter, because the Makefile appears to be generating the library name incorrectly. It is placing the version number before the ".so". Which my understanding means nothing will link against it because to link against that you would need to call GCC with `-lfdt-1.7.0-rc1`.

Makefile
```
$ ls -l libfdt/*.so*
libfdt/libfdt-1.7.0-rc1.so
libfdt/libfdt.so.1 -> libfdt-1.7.0-rc1.so
```

Meson
```
$ ls -l build/libfdt/*.so*
build/libfdt/libfdt.so -> libfdt.so.1
build/libfdt/libfdt.so.1 -> libfdt.so.1.7.0
build/libfdt/libfdt.so.1.7.0
```

Assuming the Makefile version is broken, I could change the Makefile to generate the correct `.so.1.7.0` file, and then strip off the `-rc1` assuming it doesn't matter for the library filename.

dgibson: > Meson uses the major version number for the soname, so they end up the same. Below is examining the soname with an "-rc1" release candidate added to the VERSION.txt.

Ok, good to know.

> **However** the final shared library names are different. But in this case, I believe it does not matter, because the Makefile is generating the library name incorrectly. It is placing the version number before the ".so". Which my understanding means nothing will link against it because to do so you would need to call GCC with `-lfdt-1.7.0-rc1`.

There actually seem to be two different conventions: one is `libname.so.x.y.z` with a link from the soname.  The other is `libname-x.y.z.so` but again with links from the soname.  I believe the dynamic linker finds the library via the soname link so it doesn't really care.  Examples from my laptop's `/usr/lib64`:

```
$ ls -l libz.*
lrwxrwxrwx  1 root root     14 Jan 21  2023 libz.so -> libz.so.1.2.13*
lrwxrwxrwx. 1 root root     14 Jan 21  2023 libz.so.1 -> libz.so.1.2.13*
-rwxr-xr-x. 1 root root 107416 Jan 21  2023 libz.so.1.2.13*
$ ls -l libelf*
-rwxr-xr-x 1 root root 116672 Jun 22 10:00 libelf-0.189.so*
lrwxrwxrwx 1 root root     15 Jun 22 10:00 libelf.so.1 -> libelf-0.189.so*
```

AIUI, `ldconfig` is responsible for maintaining the links either way.

> Makefile
> 
> ```
> $ ls -l libfdt/*.so*
> libfdt/libfdt-1.7.0-rc1.so
> libfdt/libfdt.so.1 -> libfdt-1.7.0-rc1.so
> ```
> 
> Meson
> 
> ```
> $ ls -l build/libfdt/*.so*
> build/libfdt/libfdt.so -> libfdt.so.1
> build/libfdt/libfdt.so.1 -> libfdt.so.1.7.0
> build/libfdt/libfdt.so.1.7.0
> ```

Right.  I believe it's only the links that matter to the dynamic linker, so I don't know that you can really call the Makefile version "wrong" per se.

> Assuming the Makefile version is broken, I could change the Makefile to generate the correct `.so.1.7.0` file, and then strip off the `-rc1` assuming it doesn't matter for the library filename.

 I'd be happy enough to standardize on the `libname.so.x.y.z` format - it does seem to be a little more common than the `libname-x.y.z.so` format.

I'm a little bit hesitant about dropping the `-rc1` or other extraversion stuff.  IIUC it should already not matter to the dynamic linker, and provides a little extra documentation to someone looking at the files.  That said, it's not really that important, and we've rarely used `EXTRAVERSION` anyway.

blmaier: > There actually seem to be two different conventions: one is `libname.so.x.y.z` with a link from the soname. The other is `libname-x.y.z.so` but again with links from the soname. I believe the dynamic linker finds the library via the soname link so it doesn't really care. Examples from my laptop's `/usr/lib64`:

Ah I think I understand, so ldconfig appears to generate the libfdt.so.1 symlink based on the libraries ELF SONAME section. And GCC appears to only look for the libfdt.so.1 symlink. So the library file could be named anything as long as ldconfig can find it. Thanks for the clarification!

> Right. I believe it's only the links that matter to the dynamic linker, so I don't know that you can really call the Makefile version "wrong" per se.

Agreed based on new info :)

> I'd be happy enough to standardize on the `libname.so.x.y.z` format - it does seem to be a little more common than the `libname-x.y.z.so` format.
>
> I'm a little bit hesitant about dropping the `-rc1` or other extraversion stuff. IIUC it should already not matter to the dynamic linker, and provides a little extra documentation to someone looking at the files. That said, it's not really that important, and we've rarely used `EXTRAVERSION` anyway.

Since it doesn't seem to matter if the X.Y.Z is before the .so, or if it has an -rcX suffix, I'm ok leaving it. The downside is that Meson enforces the libfdt.so.X.Y.Z format, so they will generate different filenames. I can push another patch to change the Makefile library name and leave the choice to you.

dgibson: > Ah I think I understand, so ldconfig appears to generate the libfdt.so.1 symlink based on the libraries ELF SONAME section. 

Right

> And GCC appears to only look for the libfdt.so.1 symlink. So the library file could be named anything as long as ldconfig can find it. Thanks for the clarification!

Well.. there's two things to consider here.  First is where gcc looks at compile time - well, strictly speaking it's `ld` at link time, but that's invoked by gcc.  Second is where the dynamic linker (aka ldso or ELF interpreter) looks for the shared library when the binary linked against libfdt is executed.

I'm pretty sure the dynamic linker just looks for the link named after the soname.  The compile time linker may a bit more complex: different library versions can be source compatible but not binary compatible.  Traditionally I think the compile time linker looked for something simply called `libwhatever.so`.  That's usually link to the latest version of the shared library.  However in some cases it can also be a linker script (`/usr/lib64/libc.so` is a linker script on most glibc based Linux distros).  I don't think we're actually creating a plain `libfdt.so` link, so `ld` must have some kind of fallback search, but I'm not sure what the rules are exactly.

> > Right. I believe it's only the links that matter to the dynamic linker, so I don't know that you can really call the Makefile version "wrong" per se.
> 
> Agreed based on new info :)
> 
> > I'd be happy enough to standardize on the `libname.so.x.y.z` format - it does seem to be a little more common than the `libname-x.y.z.so` format.
> > I'm a little bit hesitant about dropping the `-rc1` or other extraversion stuff. IIUC it should already not matter to the dynamic linker, and provides a little extra documentation to someone looking at the files. That said, it's not really that important, and we've rarely used `EXTRAVERSION` anyway.
> 
> Since it doesn't seem to matter if the X.Y.Z is before the .so, or if it has an -rcX suffix, I'm ok leaving it. The downside is that Meson enforces the libfdt.so.X.Y.Z format, so they will generate different filenames. I can push another patch to change the Makefile library name and leave the choice to you.

I think it makes sense to standardise on the format meson is enforcing.  As for the `-rcX`, I'm inclined to just drop `EXTRAVERSION` sidestepping the issue.  I don't think I've actually used it for anything for many years.

blmaier: > I'm pretty sure the dynamic linker just looks for the link named after the soname. The compile time linker may a bit more complex: different library versions can be source compatible but not binary compatible. Traditionally I think the compile time linker looked for something simply called `libwhatever.so`. That's usually link to the latest version of the shared library. However in some cases it can also be a linker script (`/usr/lib64/libc.so` is a linker script on most glibc based Linux distros). I don't think we're actually creating a plain `libfdt.so` link, so `ld` must have some kind of fallback search, but I'm not sure what the rules are exactly.

A quick inspection on my machine and yes /usr/lib/\*.so is a mix of symlinks, shared libraries, and some text like "INPUT(*.so.1...)", so there's some flexibility. I noticed Meson is generating the libfdt.so -> libfdt.so.1 symlink, so I added that to the Makefile as well.

> I think it makes sense to standardise on the format meson is enforcing. As for the `-rcX`, I'm inclined to just drop `EXTRAVERSION` sidestepping the issue. I don't think I've actually used it for anything for many years.

Sounds good, I pushed an update that changes the Makefile to match Meson, and dropped the Meson string parsing stuff assuming we won't use the -rc1 in VERSION.txt.

dgibson:
Looks good, thanks.  Merged.

PR #115, opened 2023-08-04
tests: generate dtbs in Meson build directory

blmaier:
When running under Meson, check_tests() is generating dtb build files in
the source directory. This is because dtb is named by appending
".test.dtb" to the full source file name.

Use basename to extract just the source filename and write it to the
working directory which is the build directory.

### DIFF
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -373,9 +373,10 @@ tree1_tests_rw () {
 check_tests () {
     tree="$1"
     shift
+    dtb=$(basename "$tree" .dts).test.dtb
     run_sh_test "$SRCDIR/dtc-checkfails.sh" "$@" -- -I dts -O dtb $tree
-    run_dtc_test -I dts -O dtb -o $tree.test.dtb -f $tree
-    run_sh_test "$SRCDIR/dtc-checkfails.sh" "$@" -- -I dtb -O dtb $tree.test.dtb
+    run_dtc_test -I dts -O dtb -o "$dtb" -f $tree
+    run_sh_test "$SRCDIR/dtc-checkfails.sh" "$@" -- -I dtb -O dtb "$dtb"
 }
 
 ALL_LAYOUTS="mts mst tms tsm smt stm"


tests/run_tests.sh
```diff
@@ -373,9 +377,10 @@ tree1_tests_rw () {
 check_tests () {
     tree="$1"
     shift
+    dtb=$BUILDDIR/$(basename "$tree").test.dtb
```
dgibson, line 380: I'm not quite understanding what's making `check_tests()` special here.  There are plenty of other places we build dtbs in the tests (every `run_dtc_test` for example).  In most cases, AFAICT, the approach taken is to retrieve the source from the base directory with an explicit `$SRCDIR` while using a relative path for the output, so expecting the script to be running in the build directory.

Why is that not working for these specific cases?  Or is it not working anywhere, in which case it needs to be fixed more widely as well.

dgibson (commented):
I need to understand the problem a bit better before applying this.

dgibson:
There are also some files created incorrectly under `pylibfdt/` with a meson build, but that can certainly be fixed as a separate patch.

tests/run_tests.sh
blmaier: Ah I had not connected that run_tests.sh was running in the build directory. So the problem then is just that it's creating the dtb by taking "$SRCDIR/mydevices.dts" and appending ".test.dtb" to it, so the output file ends up in the $SRCDIR.

I don't need to add a $BUILDDIR variable then, rather I can do `dtb=$(basename "$tree" .dts).test.dtb` to have the same result as the other tests.

blmaier: Pushed updated version without $BUILDDIR

blmaier:
> There are also some files created incorrectly under `pylibfdt/` with a meson build, but that can certainly be fixed as a separate patch.

I was looking at those, they are significantly more complicated to fix as they are generated in the backend of distutils and swig.

tests/run_tests.sh
dgibson: Ah, much nicer, thanks!.

dgibson:
> > There are also some files created incorrectly under `pylibfdt/` with a meson build, but that can certainly be fixed as a separate patch.
> 
> I was looking at those, they are significantly more complicated to fix as they are generated in the backend of distutils and swig.

That doesn't surprise me.  Oh well, maybe one day.

dgibson:
Merged, thanks!

PR #116, opened 2023-08-04
libfdt: add missing version symbols

blmaier:
> libfdt: add missing version symbols
> 
> These symbols were not added to the version script when they were added
> to libfdt.

> editorconfig: use tab indentation for version.lds
> 
> This file is indented with tabs, but editorconfig defaults all files to
> spaces.

Covers symbols that would be added by https://github.com/dgibson/dtc/pull/108


dgibson:
Merged, thanks.

PR #117, opened 2023-08-04
version_get.h: add missing info in Meson build

blmaier:
The Makefile calls the setlocalversion script to get the repo info, append it to the version, and add it to version_gen.h. Meson is skipping the repo info and just writing the version into version_gen.h.

The Meson vcs_tag() can call an external program to get the version, but it can't append the output to another string, so we must update setlocalversion to do it for us.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -59,6 +59,11 @@ py = py.find_installation(required: get_option('python'))
 swig = find_program('swig', required: get_option('python'))
 
 version_gen_h = vcs_tag(
+  command: [
+    'scripts/setlocalversion',
+    meson.current_source_dir(),
+    meson.project_version()
+  ],
   input: 'version_gen.h.in',
   output: 'version_gen.h',
 )
--- a/scripts/setlocalversion
+++ b/scripts/setlocalversion
@@ -3,11 +3,15 @@
 # Print additional version information for non-release trees.
 
 usage() {
-	echo "Usage: $0 [srctree]" >&2
+	echo "Usage: $0 [srctree] [version]" >&2
 	exit 1
 }
 
 cd "${1:-.}" || usage
+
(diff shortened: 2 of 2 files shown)


meson.build
```diff
@@ -59,6 +59,11 @@ py = py.find_installation(required: get_option('python'))
 swig = find_program('swig', required: get_option('python'))
 
 version_gen_h = vcs_tag(
+  command: [
+    'scripts/setlocalversion',
+    meson.current_source_dir(),
```
dgibson, line 64: If `vcs_tag` is already generating a version from the git information, what is it using `setlocalversion` for?

scripts/setlocalversion
```diff
@@ -3,12 +3,16 @@
...
 usage() {
-	echo "Usage: $0 [srctree]" >&2
+	echo "Usage: $0 [srctree] [version]" >&2
 	exit 1
 }
 
 cd "${1:-.}" || usage
 
+if [ -n "${2}" ]; then
+        printf '%s' "${2}"
```
dgibson, line 13: Uh... isn't this doing nothing but echoing back the second argument if given.  Why is that useful?

meson.build
blmaier: So, on my local build vcs_tag() was not including the git info, it was just printing the fallback, which is the project version "1.7.0". I thought that was how vcs_tag() was supposed to work, and so we would need to inject it with setlocalversion.

No, turns out this is an open bug in Meson when using git-worktrees. I tried running this again in a non-worktree and it works as expected. So this Pull Request is not needed.
https://www.github.com/mesonbuild/meson/issues/8567

dgibson: Aha, thanks for the clarification.

PR #118, opened 2023-09-02
github: add workflow for Meson builds

blmaier:
This pull request is based on #110 

This does not enable the flag "--fatal-meson-warnings" for meson-setup. There is an issue with setup on the Ubuntu builder that causes this warning.

> WARNING: Broken python installation detected. Python files installed
    by Meson might not be found by python interpreter.

The issue appears to be a false-positive on Debian-based distros, introduced in Meson 0.60.0 and fixed in 0.61.5 and 0.62.0. But the Ubuntu builder currently uses 0.61.3.
See this discussion for more detail https://gitlab.com/qemu-project/qemu/-/issues/873

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ name: Build test
       - main
 
 jobs:
-  build:
+  build-make:
     runs-on: ubuntu-latest
 
     strategy:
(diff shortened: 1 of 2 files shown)


dgibson:
Merged, thanks.

PR #124, opened 2024-02-21
libfdt: rework shared/static libraries

neheb:
Instead of creating 2 libraries manualy, just call both_libraries and link to the appropriate one as requested.

Fixes compilation when passing -Ddefault_libraries=both as the static_library name is duplicated.

ping @eli-schwartz @alyssais

### DIFF
--- a/libfdt/meson.build
+++ b/libfdt/meson.build
@@ -26,7 +26,7 @@ else
 endif
 
 link_args += version_script
-libfdt = library(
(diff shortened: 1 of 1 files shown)


eli-schwartz (approved):
This makes sense. It also allows meson to detect when the objects in both libraries are built with identical command lines and therefore build them only once.

dgibson:
Merged, thanks.

PR #125, opened 2024-03-05
fix MinGW format attribute

neheb:
Setting -D__USE_MINGW_ANSI_STDIO=1 is wrong and should not be used. MinGW internally uses a macro to select between gnu_printf and printf. Just use that instead of using a wrong format under clang backends.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -22,13 +22,6 @@ add_project_arguments(
   language: 'c'
 )
 
-if host_machine.system() == 'windows'
-  add_project_arguments(
-    '-D__USE_MINGW_ANSI_STDIO=1',
-    language: 'c'
-  )
-endif
-
 add_project_arguments(
   '-DFDT_ASSUME_MASK=' + get_option('assume-mask').to_string(),
   language: 'c'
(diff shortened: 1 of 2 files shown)


dgibson:
@elmarco you originally added that `-D` as part of the meson build.  Does this change seem reasonable to you?

elmarco:
The rationale is from commit 588a29ff2e4e9291143d837daa6b9b5f0a9fa91c. I don't know whether this has changed.

Using __MINGW_PRINTF_FORMAT makes the situation even more complex imho, as we end up with 3 format variants.

@neheb what are you solving? clang doesn't support __USE_MINGW_ANSI_STDIO ?

neheb:
clang doesn’t support gnu_printf. That define is used to solve format warnings when ms_printf is used. It’s not the correct solution.

elmarco:
as long as it compiles and run the tests, I guess it's fine. @dgibson

dgibson:
Well, the tests aren't so useful, since I'm not set up to do Windows builds.  Oh well, I don't see anything obviously wrong, so applying.

dgibson:
@neheb if you have the chance and ability to add a github job for windows builds of dtc, that would be fantastic.  I don't know enough about either gitlab actions or windows to do so.

PR #126, opened 2024-03-06
github: add windows/msys CI build

elmarco:
Sadly, tests do not compile on windows atm.

@neheb

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,12 @@ name: Build test
     branches:
       - main
 
+# ensure that the workflow is only triggered once per PR,  subsequent pushes to the PR will cancel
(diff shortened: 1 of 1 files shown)


dgibson:
A pity about the tests, but this is still way better than nothing.  Thanks, applied.

.github/workflows/build.yml
```diff
@@ -63,3 +69,45 @@ jobs:
...
       - name: Run check
         run: if ! meson test -C build; then cat build/meson-logs/testlog.txt; false; fi
+
+  build-windows:
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - { sys: mingw32 }
```
neheb, line 79: mingw32 is deprecated upstream.

dgibson: That may be, but people may still be using it, so it's still useful to test.

PR #127, opened 2024-03-18
github: enforce testing pylibfdt and yaml support

blmaier:
The Ubuntu runner was not building the yaml support as it's using Ubuntu 22 (jammy) which uses libyaml 0.2.2, but the build requires libyaml 0.2.3. Switch to Ubuntu 23 which has libyaml 0.2.5.

This was not detected by the runner as the Yaml feature defaults to "auto" which turns off if it fails to find the dependency. In the runner force yaml to enabled so if it fails to build it will trigger a build failure.

We also force python support for the same reason.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


dgibson:
Applied, thanks.

PR #129, opened 2024-03-18
meson: fix pylibfdt missing dependency on libfdt

blmaier:
The python library requires libfdt to build. This would intermittently fail depending on what order targets compiled.


dgibson:
Applied, thanks.

PR #130, opened 2024-03-18
meson: fix dependencies of tests

blmaier:
meson: fix dependencies of tests

    If the tests are run without a full compile they will fail. For example
    with the following.

    > rm -rf build/
    > meson setup build/
    > meson test -C build/

    This is because the tests rely on the devicetree tools and test
    executables.

meson: split run-tests by type

    Instead of running run-tests on all tests, split them down into the 9
    separate run-tests test types. This provides better granularity of test
    results from the Meson test harness.


dgibson:
Nice!  Applied.

PR #131, opened 2024-03-21
tests: fix tests broken under Meson

blmaier:
Tests running under Meson run from a different working directory then under Makefile. Some of these tests had not been fixed to work from a different directory because the tests were testing for an error condition which is indistinguishable from a missing file.


dgibson:
Good catch.  Applied, thanks.

PR #132, opened 2024-03-31
libfdt: overlay: Refactor overlay_fixup_phandle and update test case

Drupping:
This commit makes two main changes:
1. Refactored overlay_fixup_phandle to optimize efficiency by moving the phandle lookup logic based on label outside the overlay_fixup_one_phandle call. This avoids redundant phandle lookups when a single label is associated with multiple modifications.

2. Updated the test case for overlay_bad_fixup. Changed the target DTS from overlay_base_no_symbols.test.dtb to overlay_base_manual_symbols.test.dtb. This ensures that the test case doesn't exit prematurely due to the absence of label-linked phandle in the __symbols__ node. The update guarantees that the test case appropriately checks the validity of the fixup string linked to the label, as intended.

### DIFF
--- a/libfdt/fdt_overlay.c
+++ b/libfdt/fdt_overlay.c
@@ -315,7 +315,7 @@ static int overlay_update_local_references(void *fdto, uint32_t delta)
  * @name: Name of the property holding the phandle reference in the overlay
  * @name_len: number of name characters to consider
  * @poffset: Offset within the overlay property where the phandle is stored
- * @label: Label of the node referenced by the phandle
+ * @phandle: Phandle referencing the node
  *
  * overlay_fixup_one_phandle() resolves an overlay phandle pointing to
  * a node in the base device tree.
@@ -332,29 +332,13 @@ static int overlay_fixup_one_phandle(void *fdt, void *fdto,
 				     int symbols_off,
 				     const char *path, uint32_t path_len,
 				     const char *name, uint32_t name_len,
-				     int poffset, const char *label)
+				     int poffset, uint32_t phandle)
 {
-	const char *symbol_path;
-	uint32_t phandle;
 	fdt32_t phandle_prop;
-	int symbol_off, fixup_off;
-	int prop_len;
+	int fixup_off;
 
 	if (symbols_off < 0)
 		return symbols_off;
-
-	symbol_path = fdt_getprop(fdt, symbols_off, label,
-				  &prop_len);
-	if (!symbol_path)
-		return prop_len;
-
-	symbol_off = fdt_path_offset(fdt, symbol_path);
-	if (symbol_off < 0)
-		return symbol_off;
-
-	phandle = fdt_get_phandle(fdt, symbol_off);
-	if (!phandle)
-		return -FDT_ERR_NOTFOUND;
 
 	fixup_off = fdt_path_offset_namelen(fdto, path, path_len);
 	if (fixup_off == -FDT_ERR_NOTFOUND)
@@ -394,6 +378,10 @@ static int overlay_fixup_phandle(void *fdt, void *fdto, int symbols_off,
 	const char *value;
 	const char *label;
 	int len;
+	const char *symbol_path;
+	int prop_len;
+	int symbol_off;
+	uint32_t phandle;
 
 	value = fdt_getprop_by_offset(fdto, property,
 				      &label, &len);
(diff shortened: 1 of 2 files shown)


libfdt/fdt_overlay.c
```diff
@@ -404,6 +392,18 @@ static int overlay_fixup_phandle(void *fdt, void *fdto, int symbols_off,
...
+		return prop_len;
+	
+	symbol_off = fdt_path_offset(fdt, symbol_path);
+	if (symbol_off < 0)
+		return symbol_off;
+	
+	phandle = fdt_get_phandle(fdt, symbol_off);
+	if (!phandle)
+		return -FDT_ERR_NOTFOUND;
+
```
dgibson, line 406: This is a nice refactor, thanks!.

libfdt/fdt_overlay.c
```diff
@@ -444,8 +444,8 @@ static int overlay_fixup_phandle(void *fdt, void *fdto, int symbols_off,
 			return -FDT_ERR_BADOVERLAY;
 
 		ret = overlay_fixup_one_phandle(fdt, fdto, symbols_off,
-						path, path_len, name, name_len,
-						poffset, label);
+										path, path_len, name, name_len,
+										poffset, phandle);
```
dgibson, line 448: These lines are oddly over-indented in the revised version, can you fix this up please.

tests/run_tests.sh
```diff
@@ -269,7 +269,7 @@ libfdt_overlay_tests () {
     for test in $BAD_FIXUP_TREES; do
 	tree="overlay_bad_fixup_$test"
 	run_dtc_test -I dts -O dtb -o $tree.test.dtb "$SRCDIR/$tree.dts"
-	run_test overlay_bad_fixup overlay_base_no_symbols.test.dtb $tree.test.dtb
+	run_test overlay_bad_fixup overlay_base_manual_symbols.test.dtb $tree.test.dtb
```
dgibson, line 272: This change is unrelated, logically or code-wise to the other one.  Can you please make them separate patches / commits (they can still go in the same PR on github).

dgibson (changes requested):
Sorry it's taken me so long to look at these.  The idea of the change looks good, but there are a couple of small nits I'd like improved before merge.

tests/run_tests.sh
Drupping: The test case update is necessary due to the refactoring, as it's essential for the tests to pass with the modified code logic. Specifically, the change prevents premature test exits due to the updated phandle handling. For the detailed rationale, please refer to point 2 in my original commit message. Nonetheless, I've split the commits for clarity as you suggested.

dgibson: Sorry, I missed that.  I had the impression that the first patch didn't make any behavioural change, just optimised the existing behaviour.  Since it clearly does, I need to consider more carefully if this is the right idea.

I'm having trouble understanding point 2 from your commit message.  Can you clarify for me why the behaviour is different with the overlay change?

Drupping: To clarify point two, the key difference between `overlay_base_manual_symbols.test.dtb` and `overlay_base_no_symbols.test.dtb` is that the former includes a `__symbols__` node which is crucial for our updated logic.

I'd like to draw attention to this code segment:

```c
symbol_off = fdt_path_offset(fdt, symbol_path);
if (symbol_off < 0)
	return symbol_off;
```
This block's role is to locate the `__symbols__` node in the base fdt. Before refactoring, the code checked the format of fixup string(like "/fragment@0:target:ab") in the fdto before looking up the `__symbols__` in the base fdt. After refactoring, we first check for the `__symbols__` node in the base fdt.

Therefore, post-refactor, the `__symbols__` node must exist in the base fdt to ensure the code evaluates fixup strings correctly and doesn't exit early due to missing `__symbols__`.

I hope this clarification is helpful. Let me know if further details are needed.

Drupping: Apologies for the mix-up. The code block I previously posted was incorrect. Here’s the correct one.
```c
/* And base DTs without symbols */
symbols_off = fdt_path_offset(fdt, "/__symbols__");
if ((symbols_off < 0 && (symbols_off != -FDT_ERR_NOTFOUND)))
	return symbols_off;
```
Also, my earlier explanation was not entirely accurate. Simply put, we need the base fdt to contain a `__symbols__` node so that the test case will process the logic for checking the fixup string format. Without this, the test would prematurely exit with an FDT_ERR_NOTFOUND error at ```symbol_path = fdt_getprop(fdt, symbols_off, label, &prop_len);```

dgibson: Ah, I see.  Previously the generated trees here had two defects: first they were missing the symbols, and second the actual defect we introduced deliberately to test.  With your other change we're now failing on that first defect instead of the second one we're trying to test.

dgibson:
Thanks for the clarifications.  I've now merged this - I reversed the order of the two patches so that there isn't a temporary test failure when bisecting.

PR #133, opened 2024-04-08
tests: Remove two_roots and named_root from LIBTREE_TESTS_L in Makefi…

Drupping:
…le.tests.

These two binaries are not produced;
two_roots.dtb and named_root.dtb are instead generated in TESTS_TREES. Redundant file entries eliminated.

### DIFF
--- a/tests/Makefile.tests
+++ b/tests/Makefile.tests
@@ -32,8 +32,7 @@ LIB_TESTS_L = get_mem_rsv \
 	fs_tree1
 LIB_TESTS = $(LIB_TESTS_L:%=$(TESTS_PREFIX)%)
 
-LIBTREE_TESTS_L = truncated_property truncated_string truncated_memrsv \
-	two_roots named_root
+LIBTREE_TESTS_L = truncated_property truncated_string truncated_memrsv
 
 LIBTREE_TESTS = $(LIBTREE_TESTS_L:%=$(TESTS_PREFIX)%)
 
@@ -44,7 +43,9 @@ endif
 
 TESTS = $(LIB_TESTS) $(LIBTREE_TESTS) $(DL_LIB_TESTS)
 
-TESTS_TREES_L = test_tree1.dtb
+TESTS_TREES_L = test_tree1.dtb bad_node_char.dtb bad_node_format.dtb \
+	bad_prop_char.dtb ovf_size_strings.dtb truncated_property.dtb \
+	truncated_string.dtb truncated_memrsv.dtb two_roots.dtb named_root.dtb
 TESTS_TREES = $(TESTS_TREES_L:%=$(TESTS_PREFIX)%)
 
 TESTS_TARGETS = $(TESTS) $(TESTS_TREES)


tests/Makefile.tests
```diff
@@ -32,8 +32,7 @@ LIB_TESTS_L = get_mem_rsv \
 	fs_tree1
 LIB_TESTS = $(LIB_TESTS_L:%=$(TESTS_PREFIX)%)
 
-LIBTREE_TESTS_L = truncated_property truncated_string truncated_memrsv \
-	two_roots named_root
+LIBTREE_TESTS_L = truncated_property truncated_string truncated_memrsv
```
dgibson, line 35: Contrary to the commit message, these files are not currently listed in `TESTS_TREES` although they should be, along with all the examples created by `dumptrees`.  Could you updated your patch to add all the missing things to `TEST_TREES` as well.

dgibson (changes requested):

dgibson:
So, there are actually three overlapping problems here.  Two are relatively simple:

1. `LIBTREE_TESTS` is supposed to list the test binaries which need to be linked against `trees.S` and there aren't test binaries for `named_root` and `two_roots`
2. `TREE_TESTS` should list all the dtbs which are generated by `dumptrees` from `trees.S` - currently they're all missing except for `test_tree1.dtb`.  Not just `named_root` and `two_roots` but all the others as well.

If you could update your patch to fix both of those, that would be appreciated.

The more complex problem is that we don't have any tests using the `named_root` and `two_roots` examples.  I'm not sure if they were just never written, or if they were written, but I forgot to commit them at some point.  Either way, I'm pretty sure they're lost now.  If we wrote testcases for these they'd most likely have the same names, and would be added again to `LIBTREE_TESTS`.

tests/Makefile.tests
Drupping: I apologize for any confusion caused by my commit message. To clarify, my intention was to convey that the executables for the two_roots and named_root test cases do not exist. This is distinct from the two_roots.dtb and named_root.dtb files, which are indeed generated and used by other tests. My update was aimed at removing references to the non-existent executables, not the .dtb files you've mentioned. Therefore, our statements do not conflict but address different parts of the testing setup.

Drupping:
1. removing two_roots and named_root from LIBTREE_TESTS_L was indeed the intention behind my initial commit.
2. I have already updated the patch to include all the dtbs generated by dumptrees from trees.S.
3. I suspect there was never an initial requirement to write tests for named_root and two_roots, given that testing for these scenarios seems to be covered in the commits with IDs starting with a2def54 and 4ca61f8. In these commits, two_roots and named_root were added to LIBTREE_TESTS_L but, this inclusion was redundant.

dgibson:
> 1. removing two_roots and named_root from LIBTREE_TESTS_L was indeed the intention behind my initial commit.

Right, that's what I thought.

>     2. I have already updated the patch to include all the dtbs generated by dumptrees from trees.S.

Thank you.

>     3. I suspect there was never an initial requirement to write tests for named_root and two_roots, given that testing for these scenarios seems to be covered in the commits with IDs starting with [a2def54](https://github.com/dgibson/dtc/commit/a2def5479950d066dcf829b8c1a22874d0aea9a4) and [4ca61f8](https://github.com/dgibson/dtc/commit/4ca61f84dc210ae78376d992c1ce6ebe40ecb5be). In these commits, two_roots and named_root were added to LIBTREE_TESTS_L but, this inclusion was redundant.

Ah, good point.  I missed the fact that those dtbs were used in the tests for the fdt_check function.

dgibson:
Merged, thanks.

There's an annoying test failure on windows, but it appears to be unrelated to this patch.

Issue #135, opened 2024-05-31
python extension rebuilt during meson's install phase


sharkcz:
Looks like the python extension is rebuilt when running meson's install command. Which is generally wrong and in a package build it means that the rebuild during installation won't use the proper compiler and linker flags from the compile phase, but rather the defaults. Running
````
git clean -xfd
meson setup . redhat-linux-build -Dstatic-build=false
meson compile -C redhat-linux-build -j 16 --verbose
rm -rf /tmp/dtc.ppc64le
DESTDIR=/tmp/dtc.ppc64le meson install -C redhat-linux-build --no-rebuild
````
produces
````
...
[dan@talos dtc]$ rm -rf /tmp/dtc.ppc64le; DESTDIR=/tmp/dtc.ppc64le /usr/bin/meson install -C redhat-linux-build --no-rebuild
Installing libfdt/libfdt.so.1.7.0 to /tmp/dtc.ppc64le/usr/local/lib64
Installing libfdt/libfdt.a to /tmp/dtc.ppc64le/usr/local/lib64
Installing convert-dtsv0 to /tmp/dtc.ppc64le/usr/local/bin
Installing dtc to /tmp/dtc.ppc64le/usr/local/bin
Installing fdtdump to /tmp/dtc.ppc64le/usr/local/bin
Installing fdtget to /tmp/dtc.ppc64le/usr/local/bin
Installing fdtput to /tmp/dtc.ppc64le/usr/local/bin
Installing fdtoverlay to /tmp/dtc.ppc64le/usr/local/bin
Installing /mnt/dan/dtc/libfdt/fdt.h to /tmp/dtc.ppc64le/usr/local/include
Installing /mnt/dan/dtc/libfdt/libfdt.h to /tmp/dtc.ppc64le/usr/local/include
Installing /mnt/dan/dtc/libfdt/libfdt_env.h to /tmp/dtc.ppc64le/usr/local/include
Installing /mnt/dan/dtc/redhat-linux-build/meson-private/libfdt.pc to /tmp/dtc.ppc64le/usr/local/lib64/pkgconfig
Installing /mnt/dan/dtc/dtdiff to /tmp/dtc.ppc64le/usr/local/bin
Installing symlink pointing to libfdt.so.1.7.0 to /tmp/dtc.ppc64le/usr/local/lib64/libfdt.so.1
Installing symlink pointing to libfdt.so.1 to /tmp/dtc.ppc64le/usr/local/lib64/libfdt.so
[...]

PR #136, opened 2024-06-23
pylibfdt/meson.build: fix Python library being rebuilt during install

blmaier:
User @sharkcz noted that when the '--quiet' flag is removed from ./setup.py, the following can be seen from the `./setup.py install` stage.

```
  Running custom install script 'dtc/g/pylibfdt/../setup.py --top-builddir \
    dtc/g/redhat-linux-build install --prefix=/usr/local --root=$DESTDIR'
  running install
  ...
  building '_libfdt' extension
  swigging dtc/g/pylibfdt/../pylibfdt/libfdt.i to \
    dtc/g/pylibfdt/../pylibfdt/libfdt_wrap.c
  swig -python -Idtc/g/pylibfdt/../libfdt -o \
    dtc/g/pylibfdt/../pylibfdt/libfdt_wrap.c dtc/g/pylibfdt/../pylibfdt/libfdt.i
  gcc -fno-strict-overflow -Wsign-compare -DDYNAMIC_ANNOTATIONS_ENABLED=1 \
    -DNDEBUG -fexceptions -fexceptions -fexceptions -fPIC -DPY_SSIZE_T_CLEAN \
    -Idtc/g/pylibfdt/../libfdt -I/usr/include/python3.12 -c \
    dtc/g/pylibfdt/../pylibfdt/libfdt_wrap.c -o \
    build/temp.linux-ppc64le-cpython-312dtc/g/pylibfdt/../pylibfdt/libfdt_wrap.o
  creating build/lib.linux-ppc64le-cpython-312
  gcc -shared build/temp.linux-ppc64le-cpython-312dtc/g/pylibfdt/../pylibfdt/libfdt_wrap.o \
    -Ldtc/g/redhat-linux-build/libfdt -L/usr/lib64 -lfdt -o \
    build/lib.linux-ppc64le-cpython-312/_libfdt.cpython-312-powerpc64le-linux-gnu.so
  copying dtc/g/pylibfdt/../pylibfdt/libfdt.py -> build/lib.linux-ppc64le-cpython-312

```
Meaning the python library is getting recompiled during the `meson install` phase. This causes build issues as Meson does not set the compiler and linker flags during the install phase.

The reason the library is getting rebuilt is during the normal build with "build_ext", the `--build-lib` flag gets passed which changes the default output build directory. But there is no equivalent option for the "install" command. Install instead looks in the default directory "./build" and so does not find the previously built library.

Since we can't fix the "install" command, drop the --build-lib flag. This causes setup.py to compile the libraries at
`<meson-build>/build/lib.linux-x86_64-cpython-312/`. We must also then fix run_tests.sh to find the library build directory as it's machine-dependent.

Fixes: #135

### DIFF
--- a/pylibfdt/meson.build
+++ b/pylibfdt/meson.build
@@ -6,7 +6,7 @@ pylibfdt = custom_target(
   input: 'libfdt.i',
   depends: libfdt,
   output: '_libfdt.so',
-  command: [setup_py, 'build_ext', '--build-lib=' + meson.current_build_dir()],
+  command: [setup_py, 'build_ext'],
   build_by_default: true,
 )
 
--- a/pylibfdt/Makefile.pylibfdt
+++ b/pylibfdt/Makefile.pylibfdt
@@ -17,7 +17,7 @@ endif
 
 $(PYMODULE): $(PYLIBFDT_srcs) $(LIBFDT_archive) $(SETUP)
 	@$(VECHO) PYMOD $@
-	$(PYTHON) $(SETUP) $(SETUPFLAGS) build_ext --build-lib=$(PYLIBFDT_dir)
+	$(PYTHON) $(SETUP) $(SETUPFLAGS) build_ext
 
 install_pylibfdt: $(PYMODULE)
 	@$(VECHO) INSTALL-PYLIB
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -18,7 +18,9 @@ if [ -n "$NO_PYTHON" ]; then
         no_python=false
     fi
 else
-    if [ -f ../pylibfdt/_libfdt.so ] || [ -f ../pylibfdt/_libfdt.cpython-3*.so ]; then
+    if [ -f ../pylibfdt/_libfdt.so ] \
+            || [ -f ../pylibfdt/_libfdt.cpython-3*.so ] \
+            || [ -f ../build/lib.*/_libfdt.cpython-3*.so ]; then
         no_python=false
     else
         no_python=true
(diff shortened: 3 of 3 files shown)


dgibson:
@blmaier thanks for the patch.  Unfortunately, something still seems to be missing.  It appears the logic you added to find the library during the tests only works on some distros when making meson builds: as you can see the github actions for alpine and fedora are failing, and likewise a local meson build fails the tests for me, because it's unable to import the libfdt module.

blmaier:
@dgibson weird I can't reproduce this locally, but I was able to fix it with the github workflow.

There does seem to be a clang64 test failing under Windows. Though it appears this is failing on 'main' so that is likely unrelated :smiley:

dgibson:
@blmaier thanks for the update

> @dgibson weird I can't reproduce this locally, but I was able to fix it with the github workflow.

Huh, odd.

> There does seem to be a clang64 test failing under Windows. Though it appears this is failing on 'main' so that is likely unrelated 😃

Yes, that's been around for a while, the reason is not obvious to me but seems to be something with the environment rather than a bug in dtc itself.  Don't really have the Windows knowledge to fix it.

pylibfdt/meson.build
```diff
@@ -6,7 +6,7 @@ pylibfdt = custom_target(
   input: 'libfdt.i',
   depends: libfdt,
   output: '_libfdt.so',
-  command: [setup_py, 'build_ext', '--build-lib=' + meson.current_build_dir()],
```
dgibson, line 9: So you remove `--build-lib` here, but it's still used in the Makefile build.  Is that intentional?

blmaier: Nope, I forgot to check the Makefiles. It has the same issue with rebuilding between a `make all` -> `make install`. Removing `--build-lib` fixes it there too. Pushed a second commit to fix Makefile.

dgibson:
Merged, thanks.

PR #137, opened 2024-06-24
tests/meson.build: fix python and yaml tests not running

blmaier:
The run-tests.sh script attempts to detect if Python and Yaml support is enabled in the build. This has caused false-negatives where it fails to find the Python library even though it was compiled into the build.

Avoid this problem altogether and always set the NO_PYTHON and NO_YAML to match if the features are enabled in Meson.


dgibson:
Good idea, merged.

PR #138, opened 2024-06-24
tests/run_tests.sh: fix Meson library path being dropped

blmaier:
Meson automatically passes in LD_LIBRARY_PATH pointing at the correct build directory for libfdt.so. So preserve LD_LIBRARY_PATH if it's already set.

### DIFF
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -55,7 +55,11 @@ fi
 if [ -z "$TEST_LIBDIR" ]; then
     TEST_LIBDIR=../libfdt
 fi
-export LD_LIBRARY_PATH="$TEST_LIBDIR"
+if [ -n "$LD_LIBRARY_PATH" ]; then
+    export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$TEST_LIBDIR"
+else
+    export LD_LIBRARY_PATH="$TEST_LIBDIR"
+fi
 
 export QUIET_TEST=1
 STOP_ON_FAIL=0


tests/run_tests.sh
```diff
@@ -55,7 +55,11 @@ fi
 if [ -z "$TEST_LIBDIR" ]; then
     TEST_LIBDIR=../libfdt
 fi
-export LD_LIBRARY_PATH="$TEST_LIBDIR"
+if [ -n "$LD_LIBRARY_PATH" ]; then
+    export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$TEST_LIBDIR"
```
dgibson, line 59: This doesn't seem quite right to me.  This code right here is attempting to find the library and point `LD_LIBRARY_PATH` at it.  If Meson has already done that, we don't need to do anything: we're kind of doubling up here taking both the Meson modified path and then adding a (wrong) guess at where the library is as well.

I think we'd be better off assuming that if `LD_LIBRARY_PATH` is set, then the caller knows what they're doing with it and leave it alone.

dgibson (changes requested):

tests/run_tests.sh
blmaier: > This doesn't seem quite right to me. This code right here is attempting to find the library and point LD_LIBRARY_PATH at it. If Meson has already done that, we don't need to do anything: we're kind of doubling up here taking both the Meson modified path and then adding a (wrong) guess at where the library is as well.

I'm not a fan of this code as it is guessing at where the files are, but it's still needed for the Makefile system.

> I think we'd be better off assuming that if LD_LIBRARY_PATH is set, then the caller knows what they're doing with it and leave it alone.

The issue is `LD_LIBRARY_PATH` could already be defined by the system or user for other libraries. So doing that would cause run_tests to fail to find the dtc library.

dgibson: Hrm, I see your point.  It's pretty nasty, but I can't quickly see a better way to do it.

dgibson:
Merged, thanks.

PR #139, opened 2024-06-27
pylibfdt/Makefile.pylibfdt: use project's flags to compile the extension

sharkcz:
Seems the project's CFLAGS are not used when compiling the python extension's C code via the setup.py script. Some default flags are used instead. Thus pass the CFLAGS explicitly. Unfortunately the SWIG generated code is not clean and requires `-Wno-error` override to compile successfully, because `-Werror` is used in the project's flags.


sharkcz:
IIRC meson managed builds are doing the right thing already.

dgibson:
Merged, thanks.

PR #140, opened 2024-06-30
meson: fix installation with meson-python

blmaier:
The meson-python backend fails to map 'dtdiff' install into Python wheels. Removing the prefix from the install_dir path allows meson-python to map dtdiff. The install_data(install_dir) documentation says "If this is a relative path, it is assumed to be relative to the prefix"[1]. So removing the prefix does not change the installation behaviour.

[1] https://mesonbuild.com/Reference-manual_functions.html#install_data

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -122,7 +122,7 @@ if get_option('tools')
 
   install_data(
     'dtdiff',
-    install_dir: get_option('prefix') / get_option('bindir'),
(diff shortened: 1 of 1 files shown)


blmaier:
This is preliminary work towards migrating the Python backend away from deprecated setuptools

dgibson:
Merged, thanks.

PR #141, opened 2024-06-30
libfdt: allow building only static or shared libraries

blmaier:
When using the meson-python backend to compile libfdt, it throws an installation error that it does not know where to install static libraries. We would disable the static libraries, but Meson is always building and installing both the shared and static libraries.

Right now we always build both shared and static libraries because our 'static-build' option needs us to link the dtc binaries against the static libfdt. So we need to make sure static is always built even if the build is configured for shared only.

So instead lets not link against the static library but link all of libfdt's objects directly. We will always have access to the library objects regardless if the build is configured for static or shared.

### DIFF
--- a/libfdt/meson.build
+++ b/libfdt/meson.build
@@ -26,7 +26,7 @@ else
 endif
 
 link_args += version_script
-libfdt = both_libraries(
+libfdt = library(
   'fdt', sources,
   version: meson.project_version(),
   link_args: link_args,
@@ -34,18 +34,24 @@ libfdt = both_libraries(
   install: true,
 )
 
-if static_build
-  link_with = libfdt.get_static_lib()
-else
-  link_with = libfdt.get_shared_lib()
-endif
-
 libfdt_inc = include_directories('.')
 
 libfdt_dep = declare_dependency(
   include_directories: libfdt_inc,
-  link_with: link_with,
+  link_with: libfdt,
 )
+
+if static_build
+  libfdt_static_dep = declare_dependency(
+    include_directories: libfdt_inc,
+  )
+  libfdt_static_objs = libfdt.extract_all_objects()
+else
+  libfdt_static_dep = declare_dependency(
+    dependencies: libfdt_dep
+  )
+  libfdt_static_objs = []
+endif
 
 install_headers(
   files(
--- a/meson.build
+++ b/meson.build
@@ -27,16 +27,11 @@ add_project_arguments(
   language: 'c'
 )
 
-if get_option('static-build')
-  static_build = true
-  extra_link_args = ['-static']
-else
-  static_build = false
-  extra_link_args = []
-endif
+static_build = get_option('static-build')
+static_kwargs = static_build ? {'static': true} : {}
 
 yamltree = 'yamltree.c'
-yaml = dependency('yaml-0.1', version: '>=0.2.3', required: get_option('yaml'), static: static_build)
+yaml = dependency('yaml-0.1', version: '>=0.2.3', required: get_option('yaml'), kwargs: static_kwargs)
 if not yaml.found()
   add_project_arguments('-DNO_YAML', language: 'c')
   yamltree = []
@@ -68,7 +63,7 @@ if get_option('tools')
   util_dep = declare_dependency(
     sources: ['util.c', version_gen_h],
     include_directories: '.',
-    dependencies: libfdt_dep
+    dependencies: libfdt_static_dep
   )
 
   lgen = generator(
@@ -92,7 +87,7 @@ if get_option('tools')
       ],
       dependencies: util_dep,
       install: true,
-      link_args: extra_link_args,
+      objects: libfdt_static_objs,
     )
   endif
 
@@ -113,11 +108,17 @@ if get_option('tools')
     ],
     dependencies: [util_dep, yaml],
     install: true,
-    link_args: extra_link_args,
+    objects: libfdt_static_objs,
   )
 
   foreach e: ['fdtdump', 'fdtget', 'fdtput', 'fdtoverlay']
-    dtc_tools += executable(e, files(e + '.c'), dependencies: util_dep, install: true, link_args: extra_link_args)
+    dtc_tools += executable(
+      e,
+      files(e + '.c'),
+      dependencies: util_dep,
+      install: true,
+      objects: libfdt_static_objs
+    )
   endforeach
 
   install_data(
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -104,14 +104,19 @@ if dl.found() and not static_build
   ]
 endif
 
-test_deps = [testutil_dep, util_dep, libfdt_dep]
+test_deps = [testutil_dep, util_dep]
 if not static_build
   test_deps += [dl]
 endif
 
 tests_exe = []
 foreach t: tests
-  tests_exe += executable(t, files(t + '.c'), dependencies: test_deps, link_args: extra_link_args)
+  tests_exe += executable(
+    t,
+    files(t + '.c'),
+    dependencies: test_deps,
+    objects: libfdt_static_objs,
+  )
 endforeach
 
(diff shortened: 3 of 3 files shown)


blmaier:
PR closed as I accidentally deleted branch, re-opening

libfdt/meson.build
```diff
@@ -26,27 +26,33 @@ else
 endif
 
 link_args += version_script
-libfdt = both_libraries(
+libfdt = library(
```
dgibson, line 29: I don't love the idea of avoiding `both_libraries` which seems to have some builtin logic to make our life easier.  I thought `both_libraries` already had support for selecting just one or the other libraro to build.

libfdt/meson.build
```diff
@@ -26,27 +26,33 @@ else
...
 
 libfdt_dep = declare_dependency(
   include_directories: libfdt_inc,
-  link_with: link_with,
+  link_with: libfdt,
 )
 
+if static_build
+  libfdt_static_dep = declare_dependency(
+    include_directories: libfdt_inc,
```
dgibson, line 46: Shouldn't the include directories me listed in both static and dynamic cases?

libfdt/meson.build
```diff
@@ -26,27 +26,33 @@ else
...
+  )
+  libfdt_static_objs = libfdt.extract_all_objects()
```
dgibson, line 48: IIUC, this is doing static linking by individually linking against every constituent `.o` file.  I guess that will work, but it seems convoluted, and I think we still want the ability to build a `.a` static library.

dgibson (approved):
Sorry it's taken me so long to look at this.

libfdt/meson.build
baloo: That fixes the static builds of `dtc` on NixOS.
The patch also works for dynamic builds.

I'm not sure I understand how `both_libraries` is supposed to work though.

blmaier: So the main difference between `library()` and `both_libraries()` is that `both_libraries()` _always_ builds both the static and dynamic versions of a library. Ultimately that is what I run into problems with, meson-python tries to map all the build artifacts into a Python package, but a static archive doesn't map. I'm guessing NixOS has a similar issue that it doesn't like the static archive.

One reason to use `both_libraries()` is it makes it simpler to force Meson to statically link your binaries. Since it always builds a static archive, `both_libraries()` returns the `both_libs` type which lets you do `lib.get_static_lib()` to get a direct handle. That is used later in this file to change which library is linked based on the custom `static-build` option. With just `library()` you can't access the type, and Meson decides whether to use the static or shared library.

Meson does provide some core options that let the builder influence the library type. `--default-library {shared,static,both}` and `--prefer-static`, but neither of them let you enforce a static build. I'm not sure what the value is of an option to _require_ static build, but it does complicate things.

`library()` function: https://mesonbuild.com/Reference-manual_functions.html#library
`both_library()` function: https://mesonbuild.com/Reference-manual_functions.html#both_libraries
`--default-library/--prefer-static`: https://mesonbuild.com/Builtin-options.html#core-options

libfdt/meson.build
blmaier: In the dynamic case it uses the `libfdt_dep` which is itself a dependency that has already included `libfdt_inc`.

libfdt/meson.build
blmaier: I wasn't happy with this approach either, it was the only way I could think to allow for `static-build=true` to require static linking, while also not requiring Meson compiles a static archive when `static-build=false`.

Switching from `both_libraries()` to `library()` doesn't stop us from building a `.a` static archive. It just changes it so Meson won't always build it, but instead build one or the other depending on the build configuration.

blmaier: Looking over this with fresh eyes, I think a much easier solution would be to switch between `library()` and `static_library()`. I pushed an update with that change.

eli-schwartz: This confuses me a bit... if you want a static build, you could do this:

```
meson setup builddir/ -Dc_link_args='-static' --prefer-static --default-library=static
```

The first option causes the compiler to generate executables with no dynamic section, the second option causes meson to look up dependencies statically, the third option causes meson to build internal libraries statically.

All three are decisions that various projects want to make separately, unlike "build-static" which seems to do all three in one.

blmaier: I agree `build-static` is redundant to what Meson can already do natively. I was trying to avoid removing `build-static` to not break downstreams that were using the option.

Are you suggesting to remove it entirely?

dgibson: I think that's probably a good idea, if it's redundant with what meson does automatically.  We can field complaints about breakages as they come in.

blmaier: :+1: I updated this pull request to drop build-static

libfdt/meson.build
dgibson: Even with that explanation, using a single `library` doesn't make sense to me.  It would make sense if we were building this only as an internal library in order to create the binaries that need it.  But that's not the case: we do need it as a dependency for some of the binaries here, but it's *also* a final product of the build.  We want to be able to create both static and shared version of the library as a final product, to be used by things externally.

So, if it's not possible to configure things to make `both_libraries` selectively build one or the other target, it sounds like we need separate `static_library` and `shared_library` stanzas.  By default we build both, but they could be individually disabled where necessary.  Then we'd need to configure our internal tools to have a dependency either on one or on the other depending on the preferred build type.

A case for requiring a static build is pretty easy to come up with: you want to build a binary which you can copy standalone into some foreign environment without having to also ensure that the necessary shared libraries are installed in that environment.

eli-schwartz: >  We want to be able to create both static and shared version of the library as a final product, to be used by things externally.
> 
> So, if it's not possible to configure things to make `both_libraries` selectively build one or the other target, it sounds like we need separate `static_library` and `shared_library` stanzas. By default we build both, but they could be individually disabled where necessary.


I agree, that sounds like a good idea.

As @blmaier said in the previous comment, there is a `library()` stanza that can selectively build one or the other target, or both. That is indeed the entire purpose of said function.

Meson has a *library builder* API, with five frontend hooks:

- `build_target(target_type: )` -- can specialize as one of the other 4 types using a string-dispatched value
- `library()` -- can specialize as one of the 3 types below using an option passed to `meson setup -Ddefault_library=both`
- `static_library()` -- produces a `libXXX.a` file
- `shared_library()` -- produces a `libXXX.so` file
- `both_libraries()` -- produces both `libXXX.so` and `libXXX.a`

Switching from `both_libraries()` to `library()` changes nothing other than allowing the end user performing the build to make a choice.

> Then we'd need to configure our internal tools to have a dependency either on one or on the other depending on the preferred build type.

Implemented, again, as a builtin meson option since meson 1.6.0:

https://mesonbuild.com/Builtin-options.html#details-for-default_both_libraries


So, in summary, if you accept this PR then:

- users who want both libraries can restore the status quo by configuring meson with `meson setup -Ddefault_library=both`
- users who want to individually disable static_library or shared_library can specify which one they actually want, with e.g. `-Ddefault_library=static`
- users who want both libraries, but also statically linked tools, can configure meson with `meson setup -Ddefault_library=both -Ddefault_both_libraries=static`

dgibson: Oh!  Sorry.  I misunderstood what was going on here.  I was under the impression that `library()` could build static *or* shared, but not both.  Thanks for the correction, with that information this does indeed seem like the way to go.

Given that, the only change I'd like to see is that I'd prefer `-Ddefault_library=both` to be the default behaviour with no options.

eli-schwartz: (it is cmake which famously cannot support both BUILD_SHARED_LIBS=ON and =OFF)

blmaier: @dgibson: I updated the default_options to prefer `default_library=both`

dgibson:
Applied, thanks.

PR #143, opened 2024-06-30
Pylibfdt fix python version

blmaier:
> pylibfdt: fix Python version
>       
> In commit "pylibfdt: Remove some apparently deprecated options from
> setup.py" the scm version tracking was removed. But it was not replaced
> with the fixed version. This causes the Python modules to be installed
> as version '0.0.0'.
>       
> Signed-off-by: Brandon Maier <brandon.maier@collins.com>

> pylibfdt: add VERSION.txt to Python sdist
>       
> The VERSION.txt file tells setup.py what library version to use, so we
> must include it in the source distrbution.
>       
> Signed-off-by: Brandon Maier <brandon.maier@collins.com>
>

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -9,3 +9,4 @@ include pylibfdt/libfdt.i
 include libfdt/libfdt.h
 include libfdt/fdt.h
 include libfdt/libfdt_env.h
+include VERSION.txt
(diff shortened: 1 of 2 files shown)


dgibson:
Merged, thanks.

PR #146, opened 2024-08-08
fdtoverlay: remove wrong singular article in a comment

marcows:


dgibson:
Well, it's grammatical now, but not accurate.  It should be "We take copies first" since we copy both the base tree and the overlay.

marcows:
Damn, I suspected there was something strange about that singular, but hadn't thought long enough about it.
Sorry for that, adapted the commit.

dgibson:
Merged, thanks.

PR #148, opened 2024-10-02
libfdt.h: whitespace consistency fixups

Villemoes:
In these cases, spaces are used for indentation/alignment while the surrounding lines use tab. Fix it up for consistency.


dgibson:
As with #149, this looks great, but I'll need a `Signed-off-by` line.

PR #149, opened 2024-10-02
libfdt.h: typo and consistency fixes

Villemoes:
Most of these are plain typos. Exceptions:

- "devicetree" is only used in that form in the grammar and in mailing list references; all other instances, in particular all in prose, use "device tree".

- I don't know what "nodeequested" was supposed to be, the sentence reads just fine without it.

- "inexistant" is a rare form of nonexistent.

### DIFF
--- a/libfdt/libfdt.h
+++ b/libfdt/libfdt.h
@@ -285,7 +285,7 @@ size_t fdt_header_size(const void *fdt);
 
 /**
  * fdt_header_size_ - internal function to get header size from a version number
- * @version: devicetree version number
+ * @version: device tree version number
  *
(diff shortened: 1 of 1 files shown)


dgibson:
Thanks!  This looks great, but I will need to request a `Signed-off-by` line before I can apply it.  See `CONTRIBUTING.md` for details.

Villemoes:
> Thanks! This looks great, but I will need to request a `Signed-off-by` line before I can apply it. See `CONTRIBUTING.md` for details.

Sorry, thought I had my setup do that automatically. Fixed and pushed, both.

PR #93, opened 2023-04-26
Some phandle related improvements and some bug fixes

ukleinek:
This series start with a few minor improvments.

The fourth patch fixes overlays overwriting phandles in the base dtb and thus breaking references there.

The last patch interprets the nodes `/__symbols` and `/__local_fixups` to restore labels and references when compiling to `dts`.

All patches were also sent to the devicetree-compiler mailing list.

PR #151, opened 2024-10-21
Restore phandles from binary representations

ukleinek:
device trees generated with options `-@` and `-L` contain information about the label names used originally and which values are phandle values. This information can be reused when compiling back to dts format to make the result much more human readable.

Also dtdiff is adapted to make use of this to generate smaller diffs without hunks e.g. for a different phandle allocation.

This superseeds pull request https://github.com/dgibson/dtc/pull/93.

### DIFF
--- a/dtc.c
+++ b/dtc.c
@@ -343,6 +343,8 @@ int main(int argc, char *argv[])
 	if (generate_symbols)
 		generate_label_tree(dti, "__symbols__", true);
 
+	local_fixup_phandles(dti, "__local_fixups__");
+
 	if (generate_fixups) {
 		generate_fixups_tree(dti, "__fixups__");
 		generate_local_fixups_tree(dti, "__local_fixups__");
--- a/dtc.c
+++ b/dtc.c
@@ -343,6 +343,7 @@ int main(int argc, char *argv[])
 	if (generate_symbols)
 		generate_label_tree(dti, "__symbols__", true);
 
+	fixup_phandles(dti, "__fixups__");
 	local_fixup_phandles(dti, "__local_fixups__");
 
 	if (generate_fixups) {
--- a/dtc.c
+++ b/dtc.c
@@ -338,6 +338,8 @@ int main(int argc, char *argv[])
 	if (auto_label_aliases)
 		generate_label_tree(dti, "aliases", false);
 
+	generate_labels_from_tree(dti, "__symbols__");
+
 	if (generate_symbols)
 		generate_label_tree(dti, "__symbols__", true);
 
--- a/livetree.c
+++ b/livetree.c
@@ -1090,6 +1090,102 @@ void generate_fixups_tree(struct dt_info *dti, const char *name)
 				      dti->dt);
 }
 
+void fixup_phandles(struct dt_info *dti, const char *name)
+{
+	struct node *an;
+	struct property *fp;
+
+	an = get_subnode(dti->dt, name);
+	if (!an)
+		return;
+
+	for_each_property(an, fp) {
+		char *fnext = fp->val.val;
+		char *fv;
+		unsigned int fl;
+
+		while ((fl = fp->val.len - (fnext - fp->val.val))) {
+			char *propname, *soffset;
+			struct node *n;
+			struct property *p;
+			long offset;
+
+			fv = fnext;
+			fnext = memchr(fv, 0, fl);
+
+			if (!fnext) {
+				if (quiet < 1)
+					fprintf(stderr, "Warning: Malformed fixup entry for label %s\n",
+						fp->name);
+				break;
+			}
+			fnext += 1;
+
+			propname = memchr(fv, ':', fnext - 1 - fv);
+			if (!propname) {
+				if (quiet < 1)
+					fprintf(stderr, "Warning: Malformed fixup entry for label %s\n",
+						fp->name);
+				continue;
+			}
+			propname++;
+
+			soffset = memchr(propname, ':', fnext - 1 - propname);
+			if (!soffset) {
+				if (quiet < 1)
+					fprintf(stderr, "Warning: Malformed fixup entry for label %s\n",
+						fp->name);
+				continue;
+			}
+			soffset++;
+
+			/*
+			 * temporarily modify the property to not have to create
+			 * a copy for the node path.
+			 */
+			*(propname - 1) = '\0';
+
+			n = get_node_by_path(dti->dt, fv);
+			if (!n && quiet < 1)
+				fprintf(stderr, "Warning: Label %s references non-existing node %s\n",
+					fp->name, fv);
+
+			*(propname - 1) = ':';
+
+			if (!n)
+				continue;
+
+			/*
+			 * temporarily modify the property to not have to create
+			 * a copy for the property name.
+			 */
+			*(soffset - 1) = '\0';
+
+			p = get_property(n, propname);
+
+			if (!p && quiet < 1)
+				fprintf(stderr, "Warning: Label %s references non-existing property %s in node %s\n",
+					fp->name, n->fullpath, propname);
+
+			*(soffset - 1) = ':';
+
+			if (!p)
+				continue;
+
+			offset = strtol(soffset, NULL, 0);
+			if (offset < 0 || offset + 4 > p->val.len) {
+				if (quiet < 1)
+					fprintf(stderr,
+						"Warning: Label %s contains invalid offset for property %s in node %s\n",
+						fp->name, p->name, n->fullpath);
+				continue;
+			}
+
+			property_add_marker(p, REF_PHANDLE, offset, fp->name);
+		}
+	}
+}
+
 void generate_local_fixups_tree(struct dt_info *dti, const char *name)
 {
 	struct node *n = get_subnode(dti->dt, name);
--- a/livetree.c
+++ b/livetree.c
@@ -1103,3 +1103,60 @@ void generate_local_fixups_tree(struct dt_info *dti, const char *name)
 					    build_and_name_child_node(dti->dt, name),
 					    dti->dt);
 }
+
+static void local_fixup_phandles_node(struct dt_info *dti, struct node *lf, struct node *n)
+{
+	struct property *lfp;
+	struct node *lfsubnode;
+
+	for_each_property(lf, lfp) {
+		struct property *p = get_property(n, lfp->name);
+		fdt32_t *offsets = (fdt32_t *)lfp->val.val;
+		size_t i;
+
+		if (!p) {
+			if (quiet < 1)
+				fprintf(stderr, "Warning: Property %s in %s referenced in __local_fixups__ missing\n",
+					lfp->name, n->fullpath);
+			continue;
+		}
+
+		/*
+		 * Each property in the __local_fixups__ tree is a concatenation
+		 * of offsets, so it must be a multiple of sizeof(fdt32_t).
+		 */
+		if (lfp->val.len % sizeof(fdt32_t)) {
+			if (quiet < 1)
+				fprintf(stderr, "Warning: property %s in /__local_fixups__%s malformed\n",
+					lfp->name, n->fullpath);
+			continue;
+		}
+
+		for (i = 0; i < lfp->val.len / sizeof(fdt32_t); i++)
+			add_phandle_marker(dti, p, dtb_ld32(offsets + i));
+	}
+
+	for_each_child(lf, lfsubnode) {
+		struct node *subnode = get_subnode(n, lfsubnode->name);
+
+		if (!subnode) {
+			if (quiet < 1)
+				fprintf(stderr, "Warning: node %s/%s referenced in __local_fixups__ missing\n",
+					lfsubnode->name, n->fullpath);
+			continue;
+		}
+
+		local_fixup_phandles_node(dti, lfsubnode, subnode);
+	}
+}
+
+void local_fixup_phandles(struct dt_info *dti, const char *name)
+{
+	struct node *an;
+
+	an = get_subnode(dti->dt, name);
+	if (!an)
+		return;
+
+	local_fixup_phandles_node(dti, an, dti->dt);
+}
--- a/livetree.c
+++ b/livetree.c
@@ -1046,6 +1046,27 @@ static void generate_local_fixups_tree_internal(struct dt_info *dti,
 		generate_local_fixups_tree_internal(dti, lfn, c);
 }
 
+void generate_labels_from_tree(struct dt_info *dti, const char *name)
+{
+	struct node *an;
+	struct property *p;
+
+	an = get_subnode(dti->dt, name);
+	if (!an)
+		return;
+
+	for_each_property(an, p) {
+		struct node *labeled_node;
+
+		labeled_node = get_node_by_path(dti->dt, p->val.val);
+		if (labeled_node)
+			add_label(&labeled_node->labels, p->name);
+		else if (quiet < 1)
+			fprintf(stderr, "Warning: Path %s referenced in property %s/%s missing",
+				p->val.val, name, p->name);
+	}
+}
+
 void generate_label_tree(struct dt_info *dti, const char *name, bool allocph)
 {
 	if (!any_label_tree(dti, dti->dt))
--- a/tests/over-determined.dts
+++ b/tests/over-determined.dts
@@ -1,0 +1,23 @@
+/dts-v1/;
+/plugin/;
+
+/ {
+	labelfornodea: nodea {
+		property = <&labelfornodea>;
+	};
+
+	nodeb {
+		property = <&nonexisting>;
+	};
+
+	__fixups__ {
+		nonexisting = "/nodeb:property:0";
+	};
+
+	__local_fixups__ {
+		nodea {
+			property = <0>;
+		};
+	};
+};
+
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -570,6 +570,10 @@ dtc_tests () {
     run_dtc_test -I dts -O dtb -o dtc_sized_cells.test.dtb "$SRCDIR/sized_cells.dts"
     run_test sized_cells dtc_sized_cells.test.dtb
 
+    run_dtc_test -I dts -O dts -o over-determined-once.test.dts "$SRCDIR/over-determined.dts"
+    run_dtc_test -I dts -O dts -o over-determined-twice.test.dts "over-determined-once.test.dts"
+    run_wrap_test cmp over-determined-once.test.dts over-determined-twice.test.dts
+
     run_dtc_test -I dts -O dtb -o dtc_extra-terminating-null.test.dtb "$SRCDIR/extra-terminating-null.dts"
     run_test extra-terminating-null dtc_extra-terminating-null.test.dtb
 
--- a/treesource.c
+++ b/treesource.c
@@ -139,26 +139,48 @@ static const char *delim_end[] = {
 	[TYPE_STRING] = "",
 };
 
+/*
+ * The invariants in the marker list are:
+ *  - offsets are non-strictly monotonically increasing
+ *  - for a single offset there is at most one type marker
+ *  - for a single offset that has both a type marker and non-type markers, the
+ *    type marker appears before the others.
+ */
+static struct marker **add_marker(struct marker **mi,
+				  enum markertype type, unsigned int offset, char *ref)
+{
+	struct marker *nm;
+
+	while (*mi && (*mi)->offset < offset)
+		mi = &(*mi)->next;
+
+	if (*mi && (*mi)->offset == offset && is_type_marker((*mi)->type)) {
+		if (is_type_marker(type))
+			return mi;
+		mi = &(*mi)->next;
+	}
+
+	if (*mi && (*mi)->offset == offset && type == (*mi)->type)
+		return mi;
+
+	nm = xmalloc(sizeof(*nm));
+	nm->type = type;
+	nm->offset = offset;
+	nm->ref = ref;
+	nm->next = *mi;
+	*mi = nm;
+
+	return &nm->next;
+}
+
 static void add_string_markers(struct property *prop)
 {
 	int l, len = prop->val.len;
 	const char *p = prop->val.val;
+	struct marker **mi = &prop->val.markers;
 
-	for (l = strlen(p) + 1; l < len; l += strlen(p + l) + 1) {
-		struct marker *m, **nextp;
-
-		m = xmalloc(sizeof(*m));
-		m->offset = l;
-		m->type = TYPE_STRING;
-		m->ref = NULL;
-		m->next = NULL;
-
-		/* Find the end of the markerlist */
-		nextp = &prop->val.markers;
-		while (*nextp)
-			nextp = &((*nextp)->next);
-		*nextp = m;
-	}
+	for (l = strlen(p) + 1; l < len; l += strlen(p + l) + 1)
+		mi = add_marker(mi, TYPE_STRING, l, NULL);
 }
 
 static enum markertype guess_value_type(struct property *prop)
--- a/treesource.c
+++ b/treesource.c
@@ -347,7 +347,10 @@ void dt_to_source(FILE *f, struct dt_info *dti)
 {
 	struct reserve_info *re;
 
-	fprintf(f, "/dts-v1/;\n\n");
+	fprintf(f, "/dts-v1/;\n");
+	if (dti->dtsflags & DTSF_PLUGIN)
+		fprintf(f, "/plugin/;\n");
+	fprintf(f, "\n");
 
 	for (re = dti->reservelist; re; re = re->next) {
 		struct label *l;
--- a/treesource.c
+++ b/treesource.c
@@ -183,6 +183,39 @@ static void add_string_markers(struct property *prop, unsigned int offset, int l
 		mi = add_marker(mi, TYPE_STRING, offset + l, NULL);
 }
 
+void add_phandle_marker(struct dt_info *dti, struct property *prop, unsigned int offset)
+{
+	cell_t phandle;
+	struct node *refn;
+	char *ref;
+
+	if (prop->val.len < offset + 4) {
+		if (quiet < 1)
+			fprintf(stderr,
+				"Warning: property %s too short to contain a phandle at offset %u\n",
+				prop->name, offset);
+
+	}
+
+	phandle = dtb_ld32(prop->val.val + offset);
+	refn = get_node_by_phandle(dti->dt, phandle);
+
+	if (!refn) {
+		if (quiet < 1)
+			fprintf(stderr,
+				"Warning: node referenced by phandle 0x%x in property %s not found\n",
+				phandle, prop->name);
+		return;
+	}
+
+	if (refn->labels)
+		ref = refn->labels->label;
+	else
+		ref = refn->fullpath;
+
+	add_marker(&prop->val.markers, REF_PHANDLE, offset, ref);
+}
+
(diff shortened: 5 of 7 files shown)


ukleinek:
A possible followup improvement would be to restore phandles from `__fixup__` .

ukleinek:
Just a quick headsup: I found a dtbo file where my code crashes with `-@ -L`. So don't merge yet please.

dtc.c
```diff
@@ -336,8 +336,10 @@ int main(int argc, char *argv[])
 	if (auto_label_aliases)
 		generate_label_tree(dti, "aliases", false);
 
-	if (generate_symbols)
+	if (generate_symbols) {
 		generate_label_tree(dti, "__symbols__", true);
+		generate_labels_from_tree(dti, "__symbols__");
```
dgibson, line 341: I'd not immediately sure if it make sense to only do this when `generate_symbols` is set.

livetree.c
```diff
@@ -1048,6 +1048,28 @@ static void generate_local_fixups_tree_internal(struct dt_info *dti,
...
+		return;
+
+	for_each_property(an, p) {
+		struct node *labeled_node;
+
+		labeled_node = get_node_by_path(dti->dt, p->val.val);
+		if (labeled_node)
+			add_label(&labeled_node->labels, p->name);
+		else if (quiet < 1)
+			fprintf(stderr, "Warning: Path %s referenced in %s missing",
```
dgibson, line 1069: I'd suggest suggest an explicit "referenced in symbol %s" to make it clearer what's going on without context.

dtc.c
```diff
@@ -344,6 +344,7 @@ int main(int argc, char *argv[])
 	if (generate_fixups) {
 		generate_fixups_tree(dti, "__fixups__");
 		generate_local_fixups_tree(dti, "__local_fixups__");
+		local_fixup_phandles(dti, "__local_fixups__");
```
dgibson, line 347: Again, not sure if it only makes sense to do this with `-L`.

livetree.c
```diff
@@ -3,6 +3,8 @@
  * (C) Copyright David Gibson <dwg@au1.ibm.com>, IBM Corporation.  2005.
  */
 
+#include <libfdt/libfdt.h>
+
```
dgibson, line 7: Oof.  We don't currently use libfdt within dtc, which is sort-of deliberate (they provide a cross-check to each other).  There's a case to be made to drop that policy and instead use libfdt for much of the flat tree logic within dtc.  However, I'd really prefer to do that as an overall change, using it for as much as we can, rather than adding small pieces of libfdt usage ad-hoc.

..also, I can't actually see what you're using this for.

livetree.c
```diff
@@ -1093,3 +1095,59 @@ void generate_local_fixups_tree(struct dt_info *dti, const char *name)
...
+		if (!p) {
+			if (quiet < 1)
+				fprintf(stderr, "Warning: Property %s in %s referenced in __local_fixups__ missing\n",
+					lfp->name, n->fullpath);
+			continue;
+		}
+
+		/*
+		 * Each property in the __local_fixup__ tree is a concatination
```
dgibson, line 1116: Nits: 'concatenation' and `__local_fixups__` .

livetree.c
```diff
@@ -1093,3 +1095,59 @@ void generate_local_fixups_tree(struct dt_info *dti, const char *name)
...
+		 */
+		if (lfp->val.len % sizeof(fdt32_t)) {
+			if (quiet < 1)
+				fprintf(stderr, "Warning: property %s in /__local_fixups__%s malformed\n",
+					lfp->name, n->fullpath);
+			continue;
+		}
+
+		for (i = 0; i < lfp->val.len; i += sizeof(fdt32_t))
+			add_phandle_marker(dti, p, fdt32_ld((fdt32_t *)(lfp->val.val + i)));
```
dgibson, line 1127: Not a big deal, but I'd prefer to use `i += 1` with the loop limit decreased correspondingly.  Then you can cast `lfp->val.val` to `fdt32_t *` once and just index into it with `i`.

livetree.c
```diff
@@ -1093,3 +1095,59 @@ void generate_local_fixups_tree(struct dt_info *dti, const char *name)
...
+	}
+
+	for_each_child(lf, lfsubnode) {
+		struct node *subnode = get_subnode(n, lfsubnode->name);
+
+		if (!subnode) {
+			if (quiet < 1)
+				fprintf(stderr, "Warning: subnode %s in %s referenced in __local_fixups__ missing\n",
```
dgibson, line 1135: For properties you do need to give the property name and path separately.  However for subnodes you might as well just stick them together "node %s/%s referenced in __local_fixups__ ...".

treesource.c
```diff
@@ -161,6 +163,36 @@ static void add_string_markers(struct property *prop)
...
+	refn = get_node_by_phandle(dti->dt, phandle);
+
+	if (refn) {
+		if (refn->labels)
+			m->ref = refn->labels->label;
+		else
+			m->ref = refn->fullpath;
+	} else if (quiet < 1) {
+		fprintf(stderr, "Warning: node referenced by phandle 0x%x in property %s not found\n",
+			phandle, prop->name);
```
dgibson, line 186: I think you do need to consider `__fixups__` as well.  Without that, this error will be triggered if you're decompiling a dtbo with any external phandle references.

dtdiff
```diff
@@ -17,7 +17,7 @@ source_and_sort () {
 	    *.dts)
 		IFORMAT=dts
 		;;
-	    *.dtb)
+	    *.dtb|*.dtbo)
```
dgibson, line 20: Even something this simple needs some sort of commit message: why is this change desirable?  (*I* can see why, but someone reading the commit log in the future might not).

dgibson (changes requested):
I think this is a great idea.  The implementation looks mostly sound but there are a few minor issues that need addressing.

dgibson:
> A possible followup improvement would be to restore phandles from `__fixup__` .

I think you do need to implement this.  For one thing processing but `__local_fixups__` but not `__fixups__` would just be surprising behaviour to a user, I think.  Secondly, it could cause some bogus errors in some cases (see detailed comments).

livetree.c
ukleinek: Maybe this is only a development relict, will try without.

ukleinek:
Do you have an idea for the unaligned phandle problem?

dgibson:
> Do you have an idea for the unaligned phandle problem?

Sorry, I'm not sure exactly what problem you mean.

dgibson:
> > > Do you have an idea for the unaligned phandle problem?
> > 
> > 
> > Sorry, I'm not sure exactly what problem you mean.
> 
> I mean the crash reported in [#151 (comment)](https://github.com/dgibson/dtc/pull/151#issuecomment-2434383526) with the follow up in the next comment.

Ah, right, sorry.

I think the basic problem is that previously the only case where we'd have phandle markers during decompile is with `-I dts -O dts`, in which case we'd also have type markers helping us figure out how to print each part of each property.  The new code is putting in phandle markers, but without type markers.  We should handle that, of course, but looks like there are some edge cases.

More specifically, it looks like `write_propval()` calls `guess_value_type()` to figure out a missing type only once for the whole property.  Having a phandle marker does tell us that those specific 4 bytes should be in integer `< ... >` context.  But we'll need to separately guess the type for each chunk between phandles.  I suspect that might be irritatingly fiddly, but I'm afraid it's going to be a prerequisite for implementing this.

livetree.c
ukleinek: I guess you expect that `p->name` is passed then for your `%s`? I think mentioning `name` is still a good idea, so I suggest to make this
```
fprintf(stderr, "Warning: Path %s referenced in property %s/%s missing",
        p->val.val, name, p->name);
```

ukleinek:
This addresses the crash mentioned in https://github.com/dgibson/dtc/pull/151#issuecomment-2434383526 and a few suggestions by @dgibson (which I marked as resolved). Still missing is restoring of phandles from `__fixups__`.

dtc.c
ukleinek: Maybe do it unconditionally and remove the `__symbols__` node if `generate_symbols` is set?

livetree.c
ukleinek: FTR: Just dropping the #include doesn't work. Have to take a look at a more appropriate time.

dtc.c
dgibson: Yes, I think that makes sense.

dtc.c
dgibson: As with `__symbols__` I think it probably makes sense to do this unconditionally, then strip it out of `-L` is not given.

dgibson (changes requested):

ukleinek:
OK, this got bigger now than I anticipated and also contains a few fixes, but I'm happy that I did it now and I can get it out of my head and happily use it.

Quick demo:
```
$ cat test.dts
/dts-v1/;
/plugin/;

/ {
	tralal = <&somelabel>, "text making the",
		<&somelabel>, "some more text";

	tralala = <&somelabel>, "text making the next label unaligned",
		<&somelabel>, "some more text";

	somelabel: somenode {
		property = <&nonexisting>;
	};
};
$ dtc -@ test.dts > test.dtb
```
Decompiling this dtb with dtc 1.7.2 gives you:
```
$ /usr/bin/dtc test.dtb
/dts-v1/;

/ {
	tralal = [00 00 00 01 74 65 78 74 20 6d 61 6b 69 6e 67 20 74 68 65 00 00 00 00 01 73 6f 6d 65 20 6d 6f 72 65 20 74 65 78 74 00];
	tralala = <0x01 0x74657874 0x206d616b 0x696e6720 0x74686520 0x6e657874 0x206c6162 0x656c2075 0x6e616c69 0x676e6564 0x00 0x1736f6d 0x65206d6f 0x72652074 0x65787400>;

	somenode {
		property = <0xffffffff>;
		phandle = <0x01>;
	};

	__symbols__ {
		somelabel = "/somenode";
	};

	__fixups__ {
		nonexisting = "/somenode:property:0";
	};

	__local_fixups__ {
		tralal = <0x00 0x14>;
		tralala = <0x00 0x29>;
	};
};
```
With the changes from this PR applied it gets:
```
$ dtc test.dtb
/dts-v1/;
/plugin/;

/ {
	tralal = <&somelabel>, "text making the", <&somelabel>, "some more text";
	tralala = <&somelabel>, "text making the next label unaligned", <&somelabel>, "some more text";

	somelabel: somenode {
		property = <&nonexisting>;
		phandle = <0x01>;
	};

	__symbols__ {
		somelabel = "/somenode";
	};

	__fixups__ {
		nonexisting = "/somenode:property:0";
	};

	__local_fixups__ {
		tralal = <0x00 0x14>;
		tralala = <0x00 0x29>;
	};
};
```
and if `test.dtb` was compiled without `-@` it uses `&{/somenode}` instead of `&somelabel`.

treesource.c
```diff
@@ -178,7 +177,9 @@ static enum markertype guess_value_type(struct property *prop)
 	}
 
 	for_each_marker_of_type(m, LABEL) {
-		if ((m->offset > 0) && (prop->val.val[m->offset - 1] != '\0'))
+		if (m->offset < offset || m->offset >= offset + len)
+			continue;
```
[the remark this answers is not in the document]
ukleinek: Either I'm not following you, or you're wrong. If there is a label in the middle of a data chunk, I consider that a boundary and guess the types of the two chunks separately. IMHO that makes sense because a label (or a phandle) is very likely a boundary for a type change.

treesource.c
```diff
@@ -139,16 +139,16 @@ static const char *delim_end[] = {
...
-	const char *p = prop->val.val;
+	int l;
+	const char *p = prop->val.val + offset;
 
 	for (l = strlen(p) + 1; l < len; l += strlen(p + l) + 1) {
 		struct marker *m, **nextp;
 
 		m = xmalloc(sizeof(*m));
-		m->offset = l;
+		m->offset = offset + l;
```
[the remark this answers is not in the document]
ukleinek: This is right in the new code. I think it was already correct before, but let's not waste time on that. Please convince you that the current version is fine.

ukleinek:
BTW, I didn't implement stripping out the `__fixups__`, `__local_fixups__` and `__symbols__` nodes. Mostly because it doesn't seem to be right for `-I dts -O dts` compilations. Instead the two fixup nodes are removed before they are generated to make sure that the information isn't duplicated.

treesource.c
```diff
@@ -347,7 +347,10 @@ void dt_to_source(FILE *f, struct dt_info *dti)
 {
 	struct reserve_info *re;
 
-	fprintf(f, "/dts-v1/;\n\n");
+	fprintf(f, "/dts-v1/;\n");
+	if (any_fixup_tree(dti, dti->dt))
+		fprintf(f, "/plugin/;\n");
+	fprintf(f, "\n");
```
dgibson, line 417: I think this is correct, but I seems like a very indirect way of diong it.  We're taking the `__fixups__` node, distributing across the tree as a bunch of fixup markers, then scanning the entire tree for fixup markers to mark it as `/plugin/`.

I think a clear way to do this would be to set the `DTSF_PLUGIN` flag on dtb input if there is a (non empty) `__fixups__` or `__local_fixups__` node, then on dts output output the `/plugin/` tag based on that flag.

tests/over-determined.dts
```diff
@@ -0,0 +1,23 @@
+/dts-v1/;
```
dgibson, line 1: This commit definitely needs a proper commit message.  What's the rationale for this test?

treesource.c
```diff
@@ -139,26 +139,49 @@ static const char *delim_end[] = {
 	[TYPE_STRING] = "",
 };
 
+/*
+ * The invariants in the marker list are:
+ *  - offsets are monotonically rising
```
dgibson, line 144: Maybe "non-strictly monotonically increasing", since adjacent offsets can be equal.

treesource.c
```diff
@@ -139,26 +139,49 @@ static const char *delim_end[] = {
...
+ *  - for a single offset there is at most one type marker
+ *  - for a single offset there is at most one non-type marker
```
dgibson, line 146: This one isn't true.
```
prop = &somenode, <&othernode>;
```
Will have the `REF_PATH` and `REF_PHANDLE` markers at the same offset, until the `REF_PATH` marker is resolved (before that it's essentially marking a region of zero length).

dgibson (commented):
I've applied the first three patches, because they're sensible fixes independent of the rest of the series.  I've made some comments on the next few, but I haven't completed a full review yet.

ukleinek:
> I've applied the first three patches, because they're sensible fixes independent of the rest of the series. I've made some comments on the next few, but I haven't completed a full review yet.

The topmost commit is also simple and orthogonal to the rest of the series. That one might be eligible for fast tracking, too.
I'll look into your other feedback later.

tests/over-determined.dts
ukleinek: I don't understand this concern. IMHO the commit message is fine. What are you missing?

treesource.c
ukleinek: I went with testing just for the existence of `__fixups__` and `__local_fixups__`. I think the effort to check for non-emptiness (which has a different semantic for the two subtrees) is overengineering that isn't worth being invested.

ukleinek:
I think I addressed all feedback. The only open discussion is if `__fixups__`, `__local_fixups__` and `__symbols__` should be dropped and under which conditions.

dgibson:
The first two patches looked fine and I was going to apply.. but the second one breaks a test case:
```
dtc -I dtb -O dts -o overlay_overlay_decompile.test.dts overlay_overlay.test.dtb:	PASS
dtc -I dts -O dtb -o overlay_overlay_decompile.test.dtb overlay_overlay_decompile.test.dts:PASS
dtbs_equal_ordered overlay_overlay.test.dtb overlay_overlay_decompile.test.dtb:	Starting testcase "./dtbs_equal_ordered", pid 98779
FAIL	Tag mismatch (1 != 2) at (672, 672)
```

It looks like the decompiled version is entirely missing the `__fixups__` and `__local_fixups__` which should be there.  I'm guessing this eventually gets fixed by the later patches which will reconstruct the necessary source level references to remake the fixups.  However, we don't like to break bisection like that, so that patch might have to be moved later to where it won't break things.

treesource.c
```diff
@@ -139,26 +139,48 @@ static const char *delim_end[] = {
...
 };
 
+/*
+ * The invariants in the marker list are:
+ *  - offsets are non-strictly monotonically increasing
+ *  - for a single offset there is at most one type marker
+ *  - for a single offset that has both a type marker and non-type markers, the
+ *    type marker appears before the others.
+ */
+static struct marker **add_marker(struct marker **mi,
```
dgibson, line 149: Is there a reason to take a general `struct marker **` here instead of a `struct property *`?

treesource.c
```diff
@@ -199,30 +196,49 @@ static enum markertype guess_value_type(struct property *prop)
...
+			m = add_marker(m, guess_value_type(prop, offset, (*m)->offset - offset),
+				       offset, NULL);
+
+			offset = (*m)->offset;
+		}
+
+		if ((*m)->type == REF_PHANDLE) {
+			m = add_marker(m, TYPE_UINT32, offset, NULL);
+			offset += 4;
+		}
```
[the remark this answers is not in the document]
dgibson: Although... that will require revising your invariant of at most one type marker per offset.

livetree.c
```diff
@@ -1090,6 +1090,107 @@ void generate_fixups_tree(struct dt_info *dti, const char *name)
...
+			 * temporally modify the property to not have to create
+			 * a copy for the property name.
+			 */
+			*(soffset - 1) = '\0';
+
+			p = get_property(n, propname);
+
+			if (!p && quiet < 1)
+				fprintf(stderr, "Warning: Label %s references non-existing property %s in node %s",
+					fp->name, n->fullpath, propname);
```
dgibson, line 1159: Another `continue;`?

livetree.c
```diff
@@ -1090,6 +1090,107 @@ void generate_fixups_tree(struct dt_info *dti, const char *name)
...
+			if (!p)
+				break;
+
+			offset = strtol(soffset, NULL, 0);
+			if (offset < 0 || offset + 4 > p->val.len) {
+				if (quiet < 1)
+					fprintf(stderr,
+						"Warning: Label %s contains invalid offset for property %s in node %s",
+						fp->name, p->name, n->fullpath);
+				break;
```
dgibson, line 1172: I think `continue;` would make more sense - if one fixup is bogus, you still parse the others.  The fixups are badly constructed either way, so I guess `break;` is ok, but as a general rule we try to keep going when we can.

Of couse using `continue;` here, as with the other places I suggested does mean changing from a `while` to a `for` and moving the update of `fl` and `fv` into the `for` clause.

treesource.c
```diff
@@ -173,6 +173,12 @@ static struct marker **add_marker(struct marker **mi,
 	return &nm->next;
 }
 
+void property_add_marker(struct property *prop,
```
dgibson, line 176: This belongs in `livetree.c`

dgibson (changes requested):

tests/run_tests.sh
```diff
@@ -570,6 +570,10 @@ dtc_tests () {
     run_dtc_test -I dts -O dtb -o dtc_sized_cells.test.dtb "$SRCDIR/sized_cells.dts"
     run_test sized_cells dtc_sized_cells.test.dtb
 
+    run_dtc_test -I dts -O dts -o over-determined-once.dts "$SRCDIR/over-determined.dts"
+    run_dtc_test -I dts -O dts -o over-determined-twice.dts "over-determined-once.dts"
+    run_wrap_test cmp over-determined-once.dts over-determined-twice.dts
```
dgibson, line 575: I generally mistrust directly comparing dts files, since it can pick up meaningless format differences.  But I guess both files you're comparing here are generated by the same dtc version, so it's probably ok.

ukleinek:
> The first two patches looked fine and I was going to apply.. but the second one breaks a test case:
> 
> ```
> dtc -I dtb -O dts -o overlay_overlay_decompile.test.dts overlay_overlay.test.dtb:	PASS
> dtc -I dts -O dtb -o overlay_overlay_decompile.test.dtb overlay_overlay_decompile.test.dts:PASS
> dtbs_equal_ordered overlay_overlay.test.dtb overlay_overlay_decompile.test.dtb:	Starting testcase "./dtbs_equal_ordered", pid 98779
> FAIL	Tag mismatch (1 != 2) at (672, 672)
> ```
> 
> It looks like the decompiled version is entirely missing the `__fixups__` and `__local_fixups__` which should be there. I'm guessing this eventually gets fixed by the later patches which will reconstruct the necessary source level references to remake the fixups. However, we don't like to break bisection like that, so that patch might have to be moved later to where it won't break things.

This commit uncovers a problem that we already have since commit 915daadbb62d ("Start with empty __local_fixups__ and __fixups__ nodes"):

    $ cat test.dts 
    /dts-v1/;
    /plugin/;
    
    / {
    	node {
    		property = <0xffffffff>, "string";
    	};
    	
    	__fixups__ {
    		somenode = "/node:property:0";
    	};
    };

    $ dtc -I dts -O dts test.dts 
    /dts-v1/;
    
    / {
    
    	node {
    		property = <0xffffffff>, "string";
    	};
    };

I'll think about a fix.

tests/run_tests.sh
```diff
@@ -570,6 +570,10 @@ dtc_tests () {
     run_dtc_test -I dts -O dtb -o dtc_sized_cells.test.dtb "$SRCDIR/sized_cells.dts"
     run_test sized_cells dtc_sized_cells.test.dtb
 
+    run_dtc_test -I dts -O dts -o over-determined-once.dts "$SRCDIR/over-determined.dts"
```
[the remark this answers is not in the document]
ukleinek: I fixed that locally here, will be better after the next force-push.

treesource.c
```diff
@@ -183,6 +183,31 @@ static void add_string_markers(struct property *prop, unsigned int offset, int l
...
+	phandle = dtb_ld32(prop->val.val + offset);
+	refn = get_node_by_phandle(dti->dt, phandle);
+
+	if (refn) {
+		if (refn->labels)
+			ref = refn->labels->label;
+		else
+			ref = refn->fullpath;
+	} else {
+		if (quiet < 1)
```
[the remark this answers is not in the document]
ukleinek: Fixed locally here, will collect a few more changes before pushing again.

livetree.c
```diff
@@ -1090,6 +1090,107 @@ void generate_fixups_tree(struct dt_info *dti, const char *name)
...
+			/*
+			 * temporally modify the property to not have to create
+			 * a copy for the node path.
+			 */
+			*(propname - 1) = '\0';
+
+			n = get_node_by_path(dti->dt, fv);
+			if (!n && quiet < 1)
+				fprintf(stderr, "Warning: Label %s references non-existing node %s",
+					fp->name, fv);
```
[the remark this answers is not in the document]
ukleinek: There is a `break;` a few lines further down. `*(propname - 1) = ':'` cannot be done before the error message is emitted.

livetree.c
```diff
@@ -1090,6 +1090,107 @@ void generate_fixups_tree(struct dt_info *dti, const char *name)
...
+static void fixup_phandles_node(struct dt_info *dti, struct node *f)
+{
+	struct property *fp;
+
+	for_each_property(f, fp) {
+		char *fv = fp->val.val;
+		unsigned int fl = fp->val.len;
+
+		while (fl) {
+			char *fend = memchr(fv, 0, fl);
```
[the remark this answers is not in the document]
ukleinek: I don't understand that comment.

treesource.c
ukleinek: Is this true with `add_marker` in `treesource.c`?

treesource.c
```diff
@@ -173,23 +173,20 @@ static struct marker **add_marker(struct marker **mi,
...
 {
-	int l, len = prop->val.len;
-	const char *p = prop->val.val;
+	int l;
+	const char *p = prop->val.val + offset;
 	struct marker **mi = &prop->val.markers;
 
 	for (l = strlen(p) + 1; l < len; l += strlen(p + l) + 1)
-		mi = add_marker(mi, TYPE_STRING, l, NULL);
+		mi = add_marker(mi, TYPE_STRING, offset + l, NULL);
```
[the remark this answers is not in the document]
dgibson: Ok, fair.

livetree.c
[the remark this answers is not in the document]
dgibson: Oh, sorry, I missed that.  Ugh, this is why I kind of hate temporarily mangling the data in place, not that I can really see a better way in this instance.

livetree.c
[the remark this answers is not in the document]
dgibson: Sorry, I started writing something in the wrong place, and for some reason github was being weird and didn't let me delete it properly.

dgibson:
> This commit uncovers a problem that we already have since commit [915daad](https://github.com/dgibson/dtc/commit/915daadbb62d68bc49c5a14a360d1c39e0131f97) ("Start with empty **local_fixups** and **fixups** nodes"):
> 
> ```
> $ cat test.dts 
> /dts-v1/;
> /plugin/;
> 
> / {
> 	node {
> 		property = <0xffffffff>, "string";
> 	};
> 	
> 	__fixups__ {
> 		somenode = "/node:property:0";
> 	};
> };
> 
> $ dtc -I dts -O dts test.dts 
> /dts-v1/;
> 
> / {
> 
> 	node {
> 		property = <0xffffffff>, "string";
> 	};
> };
> ```
> 
> I'll think about a fix.

Ouch, ok.  I missed that nasty side effect of [915daad](https://github.com/dgibson/dtc/commit/915daadbb62d68bc49c5a14a360d1c39e0131f97).  Unfortunately I don't think the fix proposed in the latest version is quite sufficient.

It's an important principle in dtc's design that it should generally be safe and lossless to convert dtb -> dts -> dtb.  Likewise it should be "lossless" to convert dts -> dts, in the sense that going directly dts -> dtb and going dts -> dts -> dtb should result in the same output.  This is exactly what those test cases are designed for.

This is supposed to be true even if the input tree contains garbage.  It needs the basic structure to be parseable, of course, but if the properties contain things that are semantically malformed, they should be passed through unchanged.  915daad broke that.

Your latest series partially fixes this, but not completely: it will still simply delete entries it doesn't understand in the input tree.  It might also re-order entries that it does understand.  I can see two ways to fully fix this:

1) Change fixup generation, so that instead of unconditionally adding the new fixup, it first checks to see if a matching fixup is already in `__fixups__` or `__local_fixups__`.  With that in place, 915daad can be reverted.  That fix could be applied immediately, then your code to understand the existing fixups could be done on top of that.  Alas, that is kind of messy and a fair bit of work.

2) This approach might be a bit easier... but I'm not sure it gets every case correct.  Instead of wiping `__fixups__` clean as a single step, you remove entries one at a time as you parse them into fixups in the live tree (so you now know they'll be re-generated).  This could still result in re-orderings which isn't great, but maybe acceptable.  I'm also not sure if it handles all breakages from 915daad that don't involve a dtb -> dts stage.

ukleinek:
FTR: I took this patch set to the mailing list as I prefer this workflow after this PR got unwieldy.
v1: https://lore.kernel.org/devicetree-compiler/cover.1755692822.git.u.kleine-koenig@baylibre.com
v2: https://lore.kernel.org/devicetree-compiler/20250919092912.663304-8-u.kleine-koenig@baylibre.com

dgibson:
Equivalent changes merged via mail.  Closing.

PR #152, opened 2024-10-21
Escape spaces in depfile with backslashes.

ColinFinck:
This matches how Linux escapes spaces in paths.
The same syntax is also used by other build tools that output depfiles, e.g. https://github.com/rust-lang/cargo/blob/edd36eba5e0d6e0cfcb84bd0cc651ba8bf5e7f83/src/cargo/core/compiler/output_depinfo.rs#L19

CC @dgibson

### DIFF
--- a/dtc.c
+++ b/dtc.c
@@ -289,7 +289,9 @@ int main(int argc, char *argv[])
 		if (!depfile)
 			die("Couldn't open dependency file %s: %s\n", depname,
 			    strerror(errno));
-		fprintf(depfile, "%s:", outname);
+
+		fprint_path_escaped(depfile, outname);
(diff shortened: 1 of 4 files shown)


dgibson:
The change itself looks good, but I'll need a `Signed-off-by` line in order to apply it.  See [CONTRIBUTING.md](https://github.com/dgibson/dtc/blob/main/CONTRIBUTING.md) for details.

ColinFinck:
@dgibson Thank you very much for the quick response! I've just amended my commit to include a `Signed-off-by` line (and even a GPG signature :))

dgibson:
Merged, thanks.

PR #153, opened 2024-10-28
fdtoverlay: provide better error message for missing `/__symbols__`

Amjad50:
This was added because trying to apply overlay on dtb without knowing a lot about the subject can be frustrating with strange error messages.

Before this, it will tell you:
`Failed to apply 'overlay.dtbo': FDT_ERR_BADOFFSET`

Which doesn't explain much if you don't know.

This new message is similar to what's shown in `u-boot` when trying to apply overlay.

Reason for this is that it took me a while to debug this and understand it was an issue with `/__symbols__`, and got that from `u-boot`, hopefully it will make debugging life easier for future people with similar issue.

### DIFF
--- a/fdtoverlay.c
+++ b/fdtoverlay.c
@@ -46,6 +46,7 @@ static void *apply_one(char *base, const char *overlay, size_t *buf_len,
 	char *tmp = NULL;
 	char *tmpo;
 	int ret;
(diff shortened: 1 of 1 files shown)


dgibson:
LGTM, merged.

PR #155, opened 2024-11-11
build: fix -Dtools=false build

elmarco:
Fixes:
tests/meson.build:107:27: ERROR: Unknown variable "util_dep"

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -61,15 +61,15 @@ version_gen_h = vcs_tag(
 subdir('libfdt')
 
 dtc_tools = []
+util_dep = declare_dependency(
+  sources: ['util.c', version_gen_h],
+  include_directories: '.',
+  dependencies: libfdt_dep
+)
+
 if get_option('tools')
   flex = find_program('flex', required: true)
   bison = find_program('bison', required: true)
-
-  util_dep = declare_dependency(
-    sources: ['util.c', version_gen_h],
-    include_directories: '.',
-    dependencies: libfdt_dep
-  )
 
   lgen = generator(
     flex,


meson.build
```diff
@@ -61,6 +61,7 @@ version_gen_h = vcs_tag(
 subdir('libfdt')
 
 dtc_tools = []
+util_dep = declare_dependency()
```
blmaier, line 64: I tried to build this change, but I get a build error because `util.h` isn't defined.
```
In file included from ../tests/testutils.c:36:
../tests/tests.h:128:10: fatal error: util.h: No such file or directory
  128 | #include "util.h"
      |          ^~~~~~~~
```

Instead, we can move the entire `util_dep` declaration here.

elmarco: yes, that works

dgibson:
Fix itself looks fine to me.  The commit message needs a little work.  It has what looks like a `Fixes:` tag, but it's neither kernel format (pointing to a commit which introduced the bug) nor github format (pointing to an issue describing the problem).  I'd appreciate a little more context about what went wrong here too.

blmaier:
@dgibson can this PR be merged?

dgibson:
I've merged this.  However, I'm not sure the fix is complete, `-Dtools=false` build still fails for me, but now it does it at the `meson test` stage instead of the `meson setup` stage.  @elmarco any ideas?

blmaier:
Some of the failing tests are because it tries to run the tool tests, that should be an easy fix.

However libfdt tests fail because some of the tests need a dtb, and it compiles them from a dts. Not sure how to fix that, it seems having tools enabled is a prerequisite for any of the tests.

PR #161, opened 2024-11-24
.github/workflows: update ubuntu runner to supported version

blmaier:
The Ubuntu runner fails with the following message

> E: The repository 'http://security.ubuntu.com/ubuntu mantic-security Release' does not have a Release file.

Ubuntu 23.10 is end-of-life as of July 2024 anyway. So switch to the latest Ubuntu tag, which is currently 24.04.

See https://discourse.ubuntu.com/t/edubuntu-23-10-has-reached-end-of-life-eol/46325

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


dgibson:
Thanks for the fix, merged.

PR #164, opened 2025-03-31
flattree: Optimize stringtable_insert()

ziyao233:
According to perf result, stringtable_insert() is one of the five hotest functions, which is obvious since it depends on a brute-force, quadratic-complexity method to deduplicate the string block and is called at creation of every property.

This patch optimizes the function in two ways,

- Replace brute-force deduplication with libc-provided memmem(), which is guaranteed to be in linear complexity on both glibc on musl libc. This brings roughly 24.6% reduction in execution time.

- An open-addressing hashtable is maintained to track strings already inserted. As property names are likely to be duplicated, this could filter out many existing strings, avoiding traversing the string block. This reduces another 1.2% of execution time.

On an i7-7200U Linux system with musl-libc, building the "dtbs" target in Linux 6.11's arm64 port takes 19.8s less in average, achieving 25.8% speed up.

### DIFF
--- a/flattree.c
+++ b/flattree.c
@@ -218,23 +218,134 @@ static struct emitter asm_emitter = {
 	.property = asm_emit_property,
 };
 
-static int stringtable_insert(struct data *d, const char *str)
+struct stringtable {
+	unsigned int len, cap;
+	struct data data;
+	int *slots;
+};
+
+/*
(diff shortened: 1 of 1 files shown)


ziyao233:
I've found some new opportunities to apply optimization and will close the PR for now. Will send it as a series later.

PR #172, opened 2025-08-06
Consider drive letters when checking for absolute paths on Windows.

ColinFinck:
This still requires you to specify paths with forward slashes instead of backslashes on Windows, due to many hardcoded checks for '/'.
Fortunately, the Windows user APIs all support forward slashes too.


dgibson:
Merged, thanks.

PR #176, opened 2026-01-06
Fix discarded const qualifiers

sgallagher:
It's unsafe to implicitly discard the const qualifier on a pointer. In overlay_fixup_phandle(), this was probably just an oversight, and making the "sep" variable a const char * is sufficient to fix it.

In create_node(), however, the "p" variable is directly modifying the buffer pointed to by "const char* node_name". To fix this, we need to actually make a duplicate of the buffer and operate on that instead.

This introduces a malloc()/free()  and an unbounded strdup() into the operation, but fdtput isn't a long-running service and the node_name argument comes directly from argv, so this shouldn't introduce a significant performance impact.

### DIFF
--- a/fdtput.c
+++ b/fdtput.c
@@ -254,19 +254,21 @@ static int create_paths(char **blob, const char *in_path)
 static int create_node(char **blob, const char *node_name)
 {
 	int node = 0;
-	char *p;
+	const char *p;
+	char *path = NULL;
 
 	p = strrchr(node_name, '/');
 	if (!p) {
 		report_error(node_name, -1, -FDT_ERR_BADPATH);
 		return -1;
 	}
-	*p = '\0';
 
 	*blob = realloc_node(*blob, p + 1);
 
 	if (p > node_name) {
-		node = fdt_path_offset(*blob, node_name);
+		path = xstrndup(node_name, (size_t)(p - node_name));
+		node = fdt_path_offset(*blob, path);
+		free(path);
 		if (node < 0) {
 			report_error(node_name, -1, node);
 			return -1;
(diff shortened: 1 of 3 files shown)


fdtput.c
```diff
@@ -255,30 +255,41 @@ static int create_node(char **blob, const char *node_name)
 {
 	int node = 0;
 	char *p;
+	char *nname;
 
-	p = strrchr(node_name, '/');
+	nname = strdup(node_name);
```
dgibson: There's no point duplicating the whole string then inserting a `\0` in place.  If you're going to copy and allocate, it would be better to locate the `/` then only `strdup()` the tail of the string.

fdtput.c
```diff
@@ -255,30 +255,41 @@ static int create_node(char **blob, const char *node_name)
...
+	if (!nname) {
```
dgibson: There's essentially no point handing `malloc()` and similar failures in short lived userspace programs like this.  Use `xstrdup()` instead.

dgibson (changes requested):
Cleaning this up is appreciated, but a couple of warts in the implementaiton.

There's a  some wider question here that I guess are out of scope for the patch, but bothers me a bit: Why does this function even need to take a string with a path, rather than just a plain node name?

sgallagher:
> Why does this function even need to take a string with a path, rather than just a plain node name?

I have no idea; my goal here wasn't to change any behavior at all, just to introduce `const` safety so it stops failing to build on Fedora. I assume something is relying on this behavior, since it's coming from the command-line, so I'm not touching that.

PR #177, opened 2026-01-14
Added empty node name check

Maldus512:
The Devicetree specification states:

```
Each node in the devicetree is named according to the following convention:
node-name@unit-address
The node-name component specifies the name of the node. It shall be 1 to 31 characters in length and consist solely of
characters from the set of characters in Table 2.1.
```
If I understand correctly the parser/lexer configuration, `dtc` considers a node name as a non-empty sequences of the allowed characters **plus** the `@` character; `unit-address` extraction is processed after parsing.
This has the side effect of considering an empty name plus an address as a valid node name, e.g.:

```dts
/dts-v1/; 

/ {
    @0 {
    };
};
```
Does not throw any errors.

I've added the `node_name_not_empty` check, verifying that the `node->basenamelen` is not zero (unless it's the root node).

### DIFF
--- a/checks.c
+++ b/checks.c
@@ -340,6 +340,14 @@ static void check_node_name_format(struct check *c, struct dt_info *dti,
 }
 ERROR(node_name_format, check_node_name_format, NULL, &node_name_chars);
 
+static void check_node_name_not_empty(struct check *c, struct dt_info *dti,
+				      struct node *node)
+{
+	if (node->basenamelen == 0 && node->parent != NULL)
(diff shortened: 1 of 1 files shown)


bjarki-andreasen (approved):

dgibson:
Change looks good, but it needs a commit message (your original description on this github ticket could be adapted), and a Signed-off-by line (see `CONTRIBUTING.md`).

dgibson:
Thanks, merged.

PR #183, opened 2026-05-19
libfdt: fix UBSAN null pointer in fdt_property()

olkur:
fdt_property() unconditionally calls memcpy(ptr, val, len) even when len is zero and val is NULL.  This is a legitimate calling convention for adding empty FDT properties such as "interrupt-controller", which carry no payload.

However, compilers that treat memcpy as nonnull on its pointer arguments will fire UBSAN before observing that len is zero.

Guard the memcpy() with a check on len so it is skipped entirely when there is no payload to copy, bringing the code in line with the nonnull contract.


dgibson:
Merged, thanks.

PR #189, opened 2026-08-06
pylibfdt: Grow the FdtSw buffer geometrically

alchark:
Every expansion copies the whole tree into a freshly allocated buffer, so growing by a fixed amount makes building a tree cost time quadratic in its size. This is especially painful when assembling larger images with the data inline, such as U-Boot's binman does for FIT images.

Grow by at least as much as the tree already holds, which is what variable sized arrays usually do specifically to avoid such excessive copying.

With this change, building a Rockchip TF-A+Falcon image whose FIT carries a 31 MiB kernel takes 33.1 s rather than 44.4 s, with binman itself down from 25.3 s to 14.0 s, as 7139 reallocations become 187. The images produced are byte-identical.

### DIFF
--- a/pylibfdt/libfdt.i
+++ b/pylibfdt/libfdt.i
@@ -857,7 +857,7 @@ class FdtSw(FdtRo):
     device tree. This will be increased automatically as needed as new items
     are added to the tree.
     """
-    INC_SIZE = 1024  # Expand size by this much when out of space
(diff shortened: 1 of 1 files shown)


dgibson:
I had a momentary misgiving about this: for some very constrained environments where libfdt is used, growing the buffer geometrically might consume too much memory.  But.. for those environments would not be where pylibfdt is used.

So, merged, thanks.

PR #190, opened 2026-08-06
pylibfdt: expose address cells, size cells and memory reservation add/delete accessors

alchark:
Currently the access to address-cells/size cells requires poking into the FdtRo._fdt private buffer. Similarly, adding or deleting memory reservations requires poking into the Fdt._fdt private buffer.

This complicates things such as inserting a /chosen node with a pre-set kernel command line and initrd location/size properties into an existing device tree (as is required for preparing a device tree blob for booting with no runtime patching by a bootloader, e.g. using U-Boot's Falcon mode), so add public accessors for these.

### DIFF
--- a/pylibfdt/libfdt.i
+++ b/pylibfdt/libfdt.i
@@ -302,6 +302,38 @@ class FdtRo(object):
         check_err(val[0], quiet)
         return val[1:]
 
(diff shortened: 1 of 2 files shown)


dgibson:
Merged, thanks (it will show as closed, not merged in github, because I merged via my private tree, not via github).

PR #192, opened 2026-08-06
pylibfdt: Replace removed SWIG Python 2 compatibility macros

nullr0ute:
SWIG 4.5.0 removed Python 2 compatibility macros (PyInt_*, PyString_*) from its runtime header pyhead.swg (see commit 79f7a2b7cb7d). Replace them with their Python 3 C API equivalents:

- PyString_FromString -> PyUnicode_FromString
- PyString_AsString -> PyBytes_AsString
- PyInt_AsLong -> PyLong_AsLong

The replacements are safe since the macros were already aliased to these exact functions in SWIG's Python 3 code path.

### DIFF
--- a/pylibfdt/libfdt.i
+++ b/pylibfdt/libfdt.i
@@ -1136,7 +1136,7 @@ typedef uint32_t fdt32_t;
 	PyObject *buff;
 
 	if ($1) {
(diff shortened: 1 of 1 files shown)


dgibson:
Merged, thanks!

PR #193, opened 2026-08-18
fstree: Add output to directory functionality

ganboing:
dtc now supports `-O <fs>`, complementing the `-I <fs>`. The /proc/device-tree style output can be useful for manipulating the dt using simple shell scripts.

### DIFF
--- a/dtc.c
+++ b/dtc.c
@@ -80,14 +80,15 @@ static const char * const usage_opts_help[] = {
 	 "\t\tdts - device tree source text\n"
 	 "\t\tdtb - device tree blob\n"
 	 "\t\tfs  - /proc/device-tree style directory",
-	"\n\tOutput file",
+	"\n\tOutput file or directory",
 	"\n\tOutput formats are:\n"
 	 "\t\tdts - device tree source text\n"
 	 "\t\tdtb - device tree blob\n"
 #ifndef NO_YAML
 	 "\t\tyaml - device tree encoded as YAML\n"
 #endif
-	 "\t\tasm - assembler source",
+	 "\t\tasm - assembler source\n"
+	 "\t\tfs  - /proc/device-tree style directory",
 	"\n\tBlob version to produce, defaults to "stringify(DEFAULT_FDT_VERSION)" (for dtb and asm output)",
 	"\n\tOutput dependency file",
 	"\n\tMake space for <number> reserve map entries (for dtb and asm output)",
@@ -354,7 +355,9 @@ int main(int argc, char *argv[])
 	if (sort)
 		sort_tree(dti);
 
-	if (streq(outname, "-")) {
+	if (streq(outform, "fs")) {
+		/* Output to directory, outf is not used */
+	} else if (streq(outname, "-")) {
 		outf = stdout;
 	} else {
 		outf = fopen(outname, "wb");
@@ -375,6 +378,8 @@ int main(int argc, char *argv[])
 		dt_to_blob(outf, dti, outversion);
 	} else if (streq(outform, "asm")) {
 		dt_to_asm(outf, dti, outversion);
+	} else if (streq(outform, "fs")) {
+		dt_to_fs(outname, dti);
 	} else if (streq(outform, "null")) {
 		/* do nothing */
 	} else {
--- a/fstree.c
+++ b/fstree.c
@@ -7,6 +7,17 @@
 
 #include <dirent.h>
 #include <sys/stat.h>
+#include <fcntl.h>
+
+static bool is_reserved_name(const char *name)
+{
+	return streq(name, ".") || streq(name, "..");
+}
+
+static bool is_safe_name(const char *name)
+{
+	return !is_reserved_name(name) && !strchr(name, '/');
+}
 
 static struct node *read_fstree(const char *dirname)
 {
@@ -24,8 +35,7 @@ static struct node *read_fstree(const char *dirname)
 	while ((de = readdir(d)) != NULL) {
 		char *tmpname;
 
-		if (streq(de->d_name, ".")
-		    || streq(de->d_name, ".."))
+		if (is_reserved_name(de->d_name))
 			continue;
 
 		tmpname = join_path(dirname, de->d_name);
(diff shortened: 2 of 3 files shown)


ganboing:
Any suggestions? Comments?

dtc.c
```diff
@@ -80,14 +80,15 @@ static const char * const usage_opts_help[] = {
 	 "\t\tdts - device tree source text\n"
 	 "\t\tdtb - device tree blob\n"
 	 "\t\tfs  - /proc/device-tree style directory",
-	"\n\tOutput file",
+	"\n\tOutput file/directory",
 	"\n\tOutput formats are:\n"
```
dgibson, line 84: I'd prefer "Output file or directory" to using a /

dtc.c
```diff
@@ -354,6 +355,11 @@ int main(int argc, char *argv[])
 	if (sort)
 		sort_tree(dti);
 
+	if (streq(outform, "fs")) {
+		dt_to_fs(outname, dti);
+		goto out_done;
+	}
+
```
dgibson: This should go in the chain of if/else statements handling all the outer output formats, rather than using a `goto`.

fstree.c
```diff
@@ -65,6 +75,44 @@ static struct node *read_fstree(const char *dirname)
...
+static void write_fstree(int dirfd, struct node *tree)
+{
+	struct property *prop;
+	struct node *child;
+	int fd, rc;
+	FILE *pfile;
+
+	for_each_property(tree, prop) {
+		if (!is_safe_name(prop->name))
+			continue;
```
dgibson, line 96: It's probably worth generating a warning message here to show that we couldn't fully convert the tree.

fstree.c
```diff
@@ -65,6 +75,44 @@ static struct node *read_fstree(const char *dirname)
...
+
+		fd = openat(dirfd, prop->name, O_WRONLY | O_CREAT | O_TRUNC, 0666);
+
```
dgibson, line 105: I think it would be safer not to overwrite in the unlikely event the file already exists.  That can be done by using `O_CREATE | O_EXCL` instead of `O_TRUNC`.

fstree.c
```diff
@@ -65,6 +75,44 @@ static struct node *read_fstree(const char *dirname)
...
+	struct property *prop;
+	struct node *child;
+	int fd, rc;
+	FILE *pfile;
+
+	for_each_property(tree, prop) {
+		if (!is_safe_name(prop->name))
+			continue;
+
+		fd = openat(dirfd, prop->name, O_WRONLY | O_CREAT | O_TRUNC, 0666);
```
dgibson: I have mixed feelings about using `openat(2)`.  I see that it simplifies things quite a bit, but before this we've mostly just used stdio functions.  It also makes error reporting trickier, see below.

fstree.c
```diff
@@ -65,6 +75,44 @@ static struct node *read_fstree(const char *dirname)
...
+
+		if (fd < 0)
+			die("Couldn't openat(%d, \"%s\"): %s\n", dirfd,
```
dgibson: A fd number isn't particularly useful in an error message.  More useful would be the complete relative path of the file you're trying to create, but at present you're avoiding constructing that using `openat()`.

fstree.c
```diff
@@ -65,6 +75,44 @@ static struct node *read_fstree(const char *dirname)
...
+			    prop->name, strerror(errno));
+		pfile = fdopen(fd, "wb");
+		if (fwrite(prop->val.val, prop->val.len, 1, pfile) != 1)
+			die("Couldn't write to \"%s\"\n", prop->name);
+		fclose(pfile);
+	}
+	for_each_child(tree, child) {
+		if (!is_safe_name(child->name))
+			continue;
```
dgibson, line 117: Again a warning message here would be useful.

fstree.c
```diff
@@ -65,6 +75,44 @@ static struct node *read_fstree(const char *dirname)
...
+
+		rc = mkdirat(dirfd, child->name, 0777);
+		if (rc < 0 && errno != EEXIST)
+			goto fail_dir;
```
dgibson: This doesn't save enough duplication to justify a `goto`.  Put a separate error message here.

Also, we expect to be creating new directories, I think we should fail on `EEXIST` rather than trying to write new files to an existing directory.

fstree.c
```diff
@@ -74,3 +122,18 @@ struct dt_info *dt_from_fs(const char *dirname)
...
 	return build_dt_info(DTSF_V1, NULL, tree, guess_boot_cpuid(tree));
 }
+
+void dt_to_fs(const char *dirname, struct dt_info *dti)
+{
+	int dirfd, rc;
+
+	rc = mkdir(dirname, 0777);
+	if (rc < 0 && errno != EEXIST)
+		goto fail;
```
dgibson: Again, I think `EEXIST` should be a failure - clobbering / modifying existing directories is a bad idea.

dgibson (changes requested):

fstree.c
ganboing: I've moved away from openat to stdio functions instead. No file would ever get overwritten, as there's a `mkdir` call in the beginning of the function, and it would fail if the directory already exists.

ganboing (commented):
I've removed `openat` calls, and use regular mkdir/fopen instead. It turned out that the msys64 environment used in the github workflow doesn't support `openat` anyway, so perhaps no point in doing it, unless we absolutely have to.

fstree.c
ganboing: The new change would bail out if the output directory already exists.

dtc.c
```diff
@@ -80,14 +80,15 @@ static const char * const usage_opts_help[] = {
 	 "\t\tdts - device tree source text\n"
 	 "\t\tdtb - device tree blob\n"
 	 "\t\tfs  - /proc/device-tree style directory",
-	"\n\tOutput file",
+	"\n\tOutput file or directory",
```
dgibson, line 83: I'd suggest phrasing this `Output file (or directory for -Ofs)` to avoid confusing people not using `-Ofs`.

fstree.c
```diff
@@ -65,6 +75,54 @@ static struct node *read_fstree(const char *dirname)
 	return tree;
 }
 
+
+
```
dgibson, line 79: Extraneous whitespace lines.

dgibson (commented):
Thanks for the revisions, this looks pretty good now.

I've made a couple of trivial comment, but the only remaining serious gap is that this needs some testing.  The obvious minimum would be to check on a variety of the existing sample trees that a  round trip dtb -> fs -> dtb doesn't change anything (using `dtbs_equal_unordered`).  Better yet would be to also check error behaviour in some edge cases like:

- dtbs with bad names (`.`, `..` or containing `/`)
- A node with a property and subnode of the same name

